PR #1, opened 2014-12-12
Added request and response logging for debugging reasons

koslibpro:
While using this proxy, even though I was able to do my job by just transferring packages through this, I could not debug anything. I needed to know the request my code was sending and the response I'd receive back, in order to check if everything was okay. And this kind of logging might be helpful anyway.

### DIFF
--- a/proxy.py
+++ b/proxy.py
@@ -300,6 +300,7 @@ class Proxy(multiprocessing.Process):
         self.server = None
         
         self.request = HttpParser()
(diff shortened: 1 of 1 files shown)


abhinavsingh:
Hi Konstantinos,

I don't think these messages add much weight even for debugging purposes. You are dumping entire body which may contain binary content (images) too. I will have to close this since this branch has also run out of HEAD.

PR #6, opened 2015-10-14
Added ability to introduce an artificial delay

jack-cox-captech:
I needed a little tool to emulate a bad connection for a specific connection, so I modified this proxy server to do the job.

### DIFF
--- a/proxy.py
+++ b/proxy.py
@@ -3,9 +3,9 @@
 """
     proxy.py
     ~~~~~~~~
-    
+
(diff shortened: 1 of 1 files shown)


abhinavsingh:
Thanks for the PR.

Such features are nice to have, however I think `proxy.py` is not the right place to have them.  I recommend to look into `netem/tc` for more accurate (and feature rich) suite of bad connection emulation.

PR #15, opened 2016-09-19
use socket.create_connection so ipv6 is transparently supported

thenewguy:
needed this to proxy services over a direct access tunnel

### DIFF
--- a/proxy.py
+++ b/proxy.py
@@ -18,7 +18,7 @@ __homepage__ = 'https://github.com/abhinavsingh/proxy.py'
 __license__ = 'BSD'
 
 import sys
-import multiprocessing
+import threading
 import datetime
 import argparse
 import logging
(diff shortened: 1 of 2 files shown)


abhinavsingh:
The diffs doesn't really make sense, indentation seems like have changed on all the lines. Can you fix that so that it's easier to see what changes are made. (Sorry for super late reply on this).

abhinavsingh:
As this PR is pretty old now _(unfair to expect you to address this again)_, I checked in the necessary changes to transparently support IPv6 https://github.com/abhinavsingh/proxy.py/commit/08da786f5297d4b9bbc3ff11a58b8d28c92127a8

Thanks for reporting this and apologies for not getting back to it earlier. I see you a first-time contributor and a lost opportunity for me to get more devs on the project.

PR #27, opened 2018-05-14
 Fix a bug when parsing chunked data

lxylxy123456:
If the data ends with the size without CRLF, this patch can prevent the parser from raising an error.

Test program:

    from proxy import *
    cp = ChunkParser()
    data1 = b'3\r\nabc\r\n4'
    data2 = b'\r\nabcd\r\n'
    cp.parse(data1)			# error
    # would be correct if do `cp.parse(data1 + data2)`

### DIFF
--- a/proxy.py
+++ b/proxy.py
@@ -403,9 +403,9 @@ class Proxy(multiprocessing.Process):
         host, port = self.server.addr if self.server else (None, None)
         if self.request.method == b"CONNECT":
             logger.info("%s:%s - %s %s:%s" % (self.client.addr[0], self.client.addr[1], self.request.method, host, port))
-        else:
+        elif self.request.method:
             logger.info("%s:%s - %s %s:%s%s - %s %s - %s bytes" % (self.client.addr[0], self.client.addr[1], self.request.method, host, port, self.request.build_url(), self.response.code, self.response.reason, len(self.response.raw)))
-    
+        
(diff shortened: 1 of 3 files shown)


abhinavsingh:
Can you also add a test case for the same?  Thank You!!!

lxylxy123456:
I'm sorry, I don't quite understand where and how I should add a test case. 
On the internet, the server should send data1 and data2 in two separate packets, and this program will have an error.

abhinavsingh:
Test cases for `ChunkParser` reside here. https://github.com/abhinavsingh/proxy.py/blob/develop/tests.py#L6

You can add your test program as a test case to verify fix addressed in this PR. It will also ensure that the logic doesn't break on future code refactorings/updates.

Run test cases using:

```$ make test```

abhinavsingh:
@lxylxy123456 Apologies for not merging your PR but I figured problem was more serious than I thought at first. The following commit addresses necessary changes and test cases for the same https://github.com/abhinavsingh/proxy.py/commit/4fde72fd106ff90d34971be3c867ec056554661f

There is still an edge case under which chunk parser can raise an error, example if chunk data is not followed by CRLF within the same packet. I have postponed the fix as I couldn't replicate such packet pattern in 2-days of running through proxy.py locally on my Macbook.

Thanks for reporting this issue.

PR #32, opened 2018-12-23
Added new option --disable-via-header to allow for disabling the Via header

### LINKED ISSUE #31: Possible option to disable Via header?
tinyproxy has the DisableViaHeader option. Could this be added to this implementation?

kylegibson:
Fixes #31

### DIFF
--- a/proxy.py
+++ b/proxy.py
@@ -423,13 +423,14 @@ class Proxy(threading.Thread):
     Accepts `Client` connection object and act as a proxy between client and server.
     """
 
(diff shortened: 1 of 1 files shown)


abhinavsingh:
Currently `--disable-via-header` is doing opposite of what is intended. Can we work on a test case probably?

1. If I run it as `./proxy.py`, I see no `Via` header.
2. When run as `./proxy.py --disable-via-header`, I see a `Via` header.

Can you verify at your end?

PR #35, opened 2019-01-16
Added --pac-file option to serve a Proxy Autoconfig file

pbi-qfs:


abhinavsingh:
Thanks for the PR.

I'm kind of holding back on this for proxy web server interface work. This PR sends back pac file for any request received by the proxy web server. Ideally, we would like to give users an option to also provide pac file server path. There are plans for a `proxy.py` APIs for runtime config updates.

See work under https://github.com/abhinavsingh/proxy.py/tree/multicore branch for reference. Once ready, I'll merge the PR and port it to refactored codebase.

PR #54, opened 2019-07-20
Log Level CLI arguments

abhinavsingh:
Previously valid values for `--log-level` argument were strictly matched with `DEBUG, INFO, WARNING, ERROR, CRITICAL`.  Now one can use both upper and lower case, as well as just the first character of the log level.  Example, following are value ways to set log level to DEBUG

```
--log-level DEBUG
--log-level debug
--log-level D
--log-level d
```

### DIFF
--- a/proxy.py
+++ b/proxy.py
@@ -6,18 +6,18 @@
 
     HTTP Proxy Server in Python.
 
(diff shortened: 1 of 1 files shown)


abhinavsingh:
Fixes #54

PR #92, opened 2019-09-25
bugfix when loading plugins

normal-cock:
The reason why `except HttpProtocolException as e` doesn't work as expected is that the way we load plugin in `proxy.py` is improper. We should not import a module itself in it, instead try to use `importlib.import_module("__main__")`.

For example, suppose we have a `test2.py ` the content of which is as following. The result of the code is `False`, and that's why  `except HttpProtocolException as e` doesn't work.
```python
# test2.py
class A:
    pass

from test2 import A as B

if __name__ == "__main__":
    print(A==B)
    exit(0)
```

### DIFF
--- a/proxy.py
+++ b/proxy.py
@@ -1391,18 +1391,16 @@ class HttpProtocolHandler(threading.Thread):
                                             'Upgraded client conn for plugin %s', str(plugin_))
                             elif isinstance(upgraded_sock, bool) and upgraded_sock:
                                 return True
+                except HttpProtocolException as e:
+                    logger.exception(
+                        'HttpProtocolException type raised', exc_info=e)
+                    response = e.response(self.request)
+                    if response:
+                        self.client.queue(response)
+                        # But is client also ready for writes?
+                        self.client.flush()
+                    return True
                 except Exception as e:
-                    if e.__class__.__name__ in (
-                            ProxyAuthenticationFailed.__name__, ProxyConnectionFailed.__name__,
-                            HttpRequestRejected.__name__):
-                        logger.exception(
-                            'HttpProtocolException type raised', exc_info=e)
-                        response = e.response(self.request)  # type: ignore
-                        if response:
-                            self.client.queue(response)
-                            # But is client also ready for writes?
-                            self.client.flush()
-                        return True
                     raise e
         return False
 
@@ -1522,7 +1520,8 @@ def load_plugins(plugins: bytes) -> Dict[bytes, List[type]]:
         if plugin == b'':
             continue
         module_name, klass_name = plugin.rsplit(DOT, 1)
-        module = importlib.import_module(text_(module_name))
+        module_name = __name__ if module_name == b'proxy' else text_(module_name)
+        module = importlib.import_module(__name__)
         klass = getattr(module, text_(klass_name))
         base_klass = inspect.getmro(klass)[::-1][2:][0]
         p[bytes_(base_klass.__name__)].append(klass)


abhinavsingh:
@normal-cock Awesome. Makes sense, this is much better.  But why have you deleted `python-package.yml` in the PR?

proxy.py
```diff
@@ -1481,7 +1479,10 @@ def load_plugins(plugins: bytes) -> Dict[bytes, List[type]]:
         if plugin == b'':
             continue
         module_name, klass_name = plugin.rsplit(DOT, 1)
-        module = importlib.import_module(text_(module_name))
+        if module_name == 'proxy':
```
abhinavsingh: `module_name == b'proxy'`

proxy.py
```diff
@@ -1481,7 +1479,10 @@ def load_plugins(plugins: bytes) -> Dict[bytes, List[type]]:
...
+            module = importlib.import_module('__main__')
+        else:
+            module = importlib.import_module(text_(module_name))
```
abhinavsingh: I'll also prefer something like:

```
module_name = '__main__` if module_name == b'proxy' else text_(module_name)
module = importlib.import_module(module_name)
```

Wdyt?

proxy.py
```diff
@@ -1350,18 +1350,16 @@ def handle_readables(self, readables: List[socket.socket]) -> bool:
                                             'Upgraded client conn for plugin %s', str(plugin_))
                             elif isinstance(upgraded_sock, bool) and upgraded_sock:
                                 return True
+                except HttpProtocolException as e:
+                    logger.exception(
+                        'HttpProtocolException type raised', exc_info=e)
+                    response = e.response(self.request)     # type: ignore
```
abhinavsingh: You can also remove `# type: ignore` now.  I had to add it because previously we were catching generic top-level `Exception`.  It shouldn't be required now as `HttpProtocolException` do have a `response` method.

proxy.py
normal-cock: I think the former is more clear although it have used more spaces.

normal-cock:
> @normal-cock Awesome. Makes sense, this is much better. But why have you deleted `python-package.yml` in the PR?

@abhinavsingh I remove `python-package.yml` because I don't think it should be a part of the project itself. Most importantly, the file makes me unable to push a new branch to my own github repository :( . I think it's better to delete `python-package.yml` in repository and keep it local by adding it in `.gitignore`. Wdyt?

abhinavsingh:
@normal-cock `python-package.yml` belongs to GitHub CI/CD integration, it's similar to `.travis.yml` file in the repository.  `python-package.yml` is part of GitHub actions which results in CI/CD run for every commit and PR.  See https://github.com/abhinavsingh/proxy.py/actions 

I am not sure but do you have to enable actions for your account?  GitHub actions is relatively new and unsure if it's available for all, but no reason why you can't check into your repository. What errors do you get when you try to commit to your repo?

normal-cock:
The error message when I keep `python-package.yml` is that:
> ! [remote rejected] develop2 -> develop2 (refusing to allow an integration to create or update .github/workflows/python-package.yml)

abhinavsingh:
> @abhinavsingh The error message when I keep `python-package.yml` is that:
> 
> > ! [remote rejected] develop2 -> develop2 (refusing to allow an integration to create or update .github/workflows/python-package.yml)

Signup for actions here https://github.com/features/actions and try again.  Github actions is a new feature from Github, its surprising that they are refusing you to push into repo.  My guess is since you don't have actions enabled, you are unable to push.

Can you try and let me know. Thanks

normal-cock:
I've enable actions and the error is still there. @abhinavsingh

abhinavsingh:
Reading the error message again `refusing to allow an integration to create or update`, looks like you cannot `create` or `update` existing Github action, unsure how you ended up deleting it in first place then :)

My guess is, start in a new branch which still contains `.github/workflows` folder, un-edited.  Then you should be able to push your edits for other files.   Can you try and confirm?  Will try to also raise this with Github, doesn't sound right to me.

normal-cock:
>start in a new branch which still contains .github/workflows folder, un-edited. Then you should be able to push your edits for other files.
The error is unchanged.

The phenomena is that only if I delete `.github/workflows` in git,  can I push new branch to my repository. That's strange.

normal-cock:
It seems that some has raise this with Github https://github.community/t5/GitHub-Actions/Refusing-to-allow-an-integration-to-create-or-update/td-p/32472.

abhinavsingh:
Thanks for the find. From that thread:

`Is there any solution to this? Forcing the contributors to use SSH is not an option.`

Did you close using `git clone https://github.com/abhinavsingh/proxy.py.git` or `git clone git@github.com:abhinavsingh/proxy.py.git`.  Does the later one help you to push?  This is indeed annoying from Github.

normal-cock:
@abhinavsingh Cool, it works after I change my origin from `https://github.com/normal-cock/proxy.py.git` to `git@github.com:normal-cock/proxy.py.git`

normal-cock:
I've change the pr. @abhinavsingh

proxy.py
abhinavsingh: Sounds good.

abhinavsingh:
> I've change the pr. @abhinavsingh

Awesome, can you make sure tests are passing.  I haven't used `importlib.import_module('__main__')` myself before and want to make sure it works as expected.  Thank you.

.travis.yml
```diff
@@ -21,6 +21,8 @@ matrix:
         - TESTING_ON_TRAVIS=1
 install:
   - pip3 install -r requirements-testing.txt
-script: python3 -m coverage run tests.py || python -m coverage run tests.py
+script: 
+  - python3 -m coverage run -m unittest -v tests.py || python -m coverage run -m unittest -v tests.py
+  - python3 -m coverage report -m || python -m coverage report -m
```
abhinavsingh: Unsure whats the intention here, we already generate coverage data and `coveralls` is consuming that `after_success`.  See https://coveralls.io/github/abhinavsingh/proxy.py for existing reports.

proxy.py
```diff
@@ -1521,8 +1519,10 @@ def load_plugins(plugins: bytes) -> Dict[bytes, List[type]]:
         plugin = plugin.strip()
         if plugin == b'':
             continue
+        # coverage test
         module_name, klass_name = plugin.rsplit(DOT, 1)
-        module = importlib.import_module(text_(module_name))
+        module_name = __name__ if module_name == b'proxy' else text_(module_name)
+        module = importlib.import_module(__name__)
```
abhinavsingh: I see tests are failing, perhaps you want to use `sys.modules[__name__]` when `module_name == b'proxy'`.  Wdyt?

normal-cock: The test is successful but coverage is decreased

.travis.yml
normal-cock: I want to see the coverage details, for example which lines is not covered exactly, but i have not found the information in coveralls.io. So I add the command temporarily.

proxy.py
abhinavsingh: Aah, I didn't realize.  Coverage is ok, I did mention exception for it here https://github.com/abhinavsingh/proxy.py#pull-request, probably there is a .coveragerc config to only fail build if coverage decreases by, say more than `>1%`.

.travis.yml
abhinavsingh: We can run `make coverage` locally to get line-by-line coverage.

Also, even if we generate a coverage on travis, does coveralls.io shows that to you?

proxy.py
normal-cock: I didn't notice it. 


>probably there is a .coveragerc config to only fail build if coverage decreases by

That would be great it there is.

Is there anything else to suggest?

.travis.yml
abhinavsingh: Also see https://coveralls.io/builds/25936980/source?filename=proxy.py where you can get line-by-line coverage report.  These were generated with existing integration IIUC.

abhinavsingh: Example https://coveralls.io/builds/25941085 this is the coverage report for the latest build of your PR.

normal-cock: The coverage locally is 89%, so I want to figure out  the difference.

> does coveralls.io shows that to you?

Yes, we can see the output in the log page.

normal-cock: > Also see https://coveralls.io/builds/25936980/source?filename=proxy.py where you can get line-by-line coverage report. These were generated with existing integration IIUC.

Got it. I'll remove the this change.

abhinavsingh: > Yes, we can see the output in the log page.

Great.  Here is a link to all historical builds https://coveralls.io/repos/272015/builds  As you can see, these line-by-line coverage reports are being generated since [day 1 of coveralls.io integration](https://coveralls.io/builds/25333644).

Hence, I don't see why we should add more commands to `.travis.yml` file.  I am unsure about what additional value does it provide that already doesn't exists :)  Wdyt?

abhinavsingh: > > Also see https://coveralls.io/builds/25936980/source?filename=proxy.py where you can get line-by-line coverage report. These were generated with existing integration IIUC.
> 
> Got it. I'll remove the this change.

No worries.  Thanks for poking me, I just discovered these myself 👍

proxy.py
abhinavsingh: Thanks for the PR.  All looks good 👍

.travis.yml
normal-cock: > I don't see why we should add more commands to .travis.yml file

You're right, I did find the line-by-line report before.

abhinavsingh:
Kindly `Squash and merge`.  Thank you.  LGTM.

proxy.py
```diff
@@ -1391,18 +1391,16 @@ def handle_readables(self, readables: List[socket.socket]) -> bool:
                                             'Upgraded client conn for plugin %s', str(plugin_))
                             elif isinstance(upgraded_sock, bool) and upgraded_sock:
                                 return True
+                except HttpProtocolException as e:
```
abhinavsingh: Thank you :)

normal-cock: You're welcome :)

PR #111, opened 2019-10-07
ssl.SSLError during handshake can teardown worker process

normal-cock:
prevent socket which is listening been close when TcpConnectionUninitializedException is raised

### DIFF
--- a/proxy.py
+++ b/proxy.py
@@ -396,12 +396,23 @@ class Worker(multiprocessing.Process):
                 except BlockingIOError:  # as e:
                     # logger.exception('BlockingIOError', exc_info=e)
                     continue
-                work = self.work_klass(
-                    fileno=conn.fileno(),
-                    addr=addr,
-                    **self.kwargs)
-                work.setDaemon(True)
-                work.start()
+                try:
+                    work = self.work_klass(
+                        fileno=conn.fileno(),
+                        addr=addr,
+                        **self.kwargs)
+                    work.setDaemon(True)
+                    work.start()
+                except ssl.SSLError as e:
+                    logger.exception('ssl.SSLError', exc_info=e)
+                except Exception as e:
+                    logger.exception('Error encountered', exc_info=e)
+                finally:
+                    try:
+                        conn.shutdown(socket.SHUT_RDWR)
+                    except Exception as e:
+                        logger.exception('Error trying to shutdown client socket', exc_info=e)
+                    conn.close()
         except KeyboardInterrupt:
             pass
         finally:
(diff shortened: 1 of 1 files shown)


normal-cock:
The abnormal case is that, when TcpConnectionUninitializedException is raised the worker process close the listening socket making the worker not work any more.

abhinavsingh:
I am wondering why is `TcpConnectionUninitializedException` being thrown here, we are just passing `fileno=conn.fileno()` to `ProtocolHandler` class, so connection is pretty much initialized.

`TcpConnectionUninitializedException` can still be thrown, for example, if client connection is closed, `_conn` object has been deleted but code is still trying to access the connection property (either to send or recv data).  In which case, I think it's a programming bug.

Do you have an example stack trace for further investigation?  Are you using `proxy.py` via a custom plugin?  Thank you.

normal-cock:
In my case, `TcpConnectionUninitializedException` have been thrown as result of the failure of  ssl handshake of when https is enabled. In this case, result of `self.optionally_wrap_socket` is `None`, so `TcpConnectionUninitializedException` is thrown.
```python
class ProtocolHandler(threading.Thread):
   def __init__
        ...
        conn = self.optionally_wrap_socket(self.fromfd(fileno))
        if conn is None:
            raise TcpConnectionUninitializedException()
``` 
I'm not using `proxy.py` via a custom plugin. I use it with a self-signed certificate.

abhinavsingh:
Makes sense, indeed ssl wrap can fail.  However, I think we can improve here a little so that we can log more helpful context.  Logging `logger.error('TcpConnectionUninitializedException while handling new connection')` isn't very clear of what actually happened.  I probably just raised `TcpConnectionUninitializedException` without thinking twice about it.

How about we remove the `except Exception` from here https://github.com/abhinavsingh/proxy.py/blob/develop/proxy.py#L1543-L1561 so that SSL handshake exception is actually propagated back to worker.

Then in the worker we can:
1) catch the exception
2) log helpful message
3) also shutdown/close client connection

Wdyt?

proxy.py
```diff
@@ -1540,25 +1547,15 @@ def optionally_wrap_socket(
...
-                return None
+            ctx = ssl.create_default_context(
+                ssl.Purpose.CLIENT_AUTH)
+            ctx.options |= ssl.OP_NO_SSLv2 | ssl.OP_NO_SSLv3 | ssl.OP_NO_TLSv1 | ssl.OP_NO_TLSv1_1
+            ctx.verify_mode = ssl.CERT_NONE
+            ctx.load_cert_chain(
+                certfile=self.config.certfile,
+                keyfile=self.config.keyfile)
+            conn = ctx.wrap_socket(conn, server_side=True)
+            return conn
```
abhinavsingh: Nit: This `return` statement can now be removed.

normal-cock:
Sounds better and I've changed the code.

proxy.py
```diff
@@ -396,12 +396,19 @@ def run(self) -> None:
...
+                        addr=addr,
+                        **self.kwargs)
+                    work.setDaemon(True)
+                    work.start()
+                except ssl.SSLError as e:
+                    logger.exception('ssl.SSLError', exc_info=e)
+                except Exception as e:
+                    logger.exception('Error encountered', exc_info=e)
+                finally:
+                    conn.close()
```
abhinavsingh: Let's also explicitly `conn.shutdown` here in a `try/except OSError` block.

PR #128, opened 2019-10-12
Invoke proxy plugin handle_request for each request in HTTP/1.1 pipeline or when TLS interception is enabled

### LINKED ISSUE #126: Is it possible to modify https requests?
I want to modify requests and response for https connections

The solution would be to use a mechanism similar to debugging proxies like Fiddler and charles proxy, where the proxy server decodes the https response -> modify it -> re-sign it using a self generated SSL certificate. of course the client have to trust this certificate in his device/browser

abhinavsingh:
This is similar to https://github.com/abhinavsingh/proxy.py/pull/125 but for Proxy plugins.  Fixes #126

### DIFF
--- a/proxy.py
+++ b/proxy.py
@@ -142,19 +142,25 @@ chunkParserStates = ChunkParserStates(1, 2, 3)
 
 HttpMethods = NamedTuple('HttpMethods', [
     ('GET', bytes),
-    ('PUT', bytes),
-    ('POST', bytes),
-    ('CONNECT', bytes),
     ('HEAD', bytes),
+    ('POST', bytes),
+    ('PUT', bytes),
     ('DELETE', bytes),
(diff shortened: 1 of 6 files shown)


abhinavsingh:
@knucker @moham96 jFYI

proxy.py
```diff
@@ -1093,24 +1107,38 @@ def name(self) -> str:
         return self.__class__.__name__      # pragma: no cover
 
     @abstractmethod
-    def before_upstream_connection(self) -> bool:
+    def before_upstream_connection(self, request: HttpParser) -> Optional[HttpParser]:
```
abhinavsingh: @knucker @moham96 Updated plugin definition, added some comments to clarify when these callbacks are invoked and how they can drive the behavior.

PR #146, opened 2019-10-28
Scaffolding

abhinavsingh:


dashboard/src/proxy.html
```diff
@@ -111,6 +112,15 @@
             </div>
         </div>
     </div>
+
+    <Section id="proxyHomeSection" class="app-tab"><h1>Home</h1></Section>
```
abhinavsingh: Since `proxyHomeSection` and other IDs are now part of individual typescript files, how about we dynamically add `<Section/>` on page load.  This way we'll be able to remove hardcoded IDs from html.

dashboard/src/proxy.ts
```diff
@@ -200,13 +72,40 @@ export class ProxyDashboard {
...
-        this.websocketApi.enableInspection()
-      } else if (activeTabId === 'proxyInspect') {
-        this.websocketApi.disableInspection()
+      //deactivateAll
+      this.home.disable();
+      this.apiDevelopment.disable();
+      this.websocketApi.disable();  
+      this.shortLinks.disable();
+      this.controls.disable();
+      this.settings.disable();
```
abhinavsingh: How about just calling `$('#app>app-tab').hide()`.  Then we at least won't need to add / remove new `this.newtab.disable()` when tabs are added / removed.

dashboard/src/settings.ts
```diff
@@ -0,0 +1,17 @@
+export class Settings {
+    private name:string = "proxySettings"
+  
+    constructor () {
+      
+    }
+ 
+
+    public enable(){
```
abhinavsingh: Let's just name these as `show / hide`.  `enable / disable` can mean different thing later.

dashboard/src/inspect.ts
```diff
@@ -0,0 +1,127 @@
+export class WebsocketApi {
+    private name:string = "proxyInspect"
```
abhinavsingh: Sorry, this file might be confusing :)  but `WebsocketApi` is core to the `ProxyDashboard` in the sense that it provides a simple interface to communicate with `dashboard.py`.  It implements all client side logic for every API provided by `dashboard.py` over websockets.

This file doesn't tie to `proxyInspect` tab actually.  Yes, it does provide a `enable_inspection` / `disable_inspection` commands, but they are just methods exposed by the backend.  Example, `WebsocketApi` file will not be responsible for painting the `Inspect Traffic` tab.

Lets create a separate file for inspect traffic tab :)

dashboard/src/proxy.ts
```diff
@@ -200,13 +72,40 @@ export class ProxyDashboard {
...
+
+      switch(clickedTabId) {
+        case "proxyHome":
```
abhinavsingh: Let's preserve this comment from previous version of file:

`// TODO: Tab ids shouldn't be hardcoded.`

Ideally, at some point, we want IDs in a single place so that refactoring becomes easy.  Keeping IDs within individual file is a good idea.

dashboard/src/proxy.html
```diff
@@ -60,6 +60,7 @@
 </nav>
 
 <div id="app">
+    
     <div id="api-development" class="proxy-data">
```
abhinavsingh: Ok to remove this now, since we already have `proxyApiDevelopmentSection` below

abhinavsingh:
We'll open separate PR, since now we have tons of conflicts due to merge of `stablize` into `develop`.  Closing this PR

PR #159, opened 2019-11-06
Add non-blocking embedded mode feature

abhinavsingh:
See updated `README.md` for usage info.

### DIFF
--- a/README.md
+++ b/README.md
@@ -753,11 +753,11 @@ if __name__ == '__main__':
     # ... your logic here ...
 ```
 
-`start` is simply a context manager which:
+Note that:
 
-1. Starts `proxy.py`.
+1. `start` is simply a context manager.
 2. Is similar to calling `main` except `start` won't block.
-3. Will automatically shut down `proxy.py` when your logic finishes execution.
+3. It automatically shut down `proxy.py`.
 
 Plugin Developer and Contributor Guide
 ======================================
--- a/README.md
+++ b/README.md
@@ -55,6 +55,8 @@ Table of Contents
 * [End-to-End Encryption](#end-to-end-encryption)
 * [TLS Interception](#tls-interception)
 * [Embed proxy.py](#embed-proxypy)
+    * [Blocking Mode](#blocking-mode)
+    * [Non-blocking Mode](#non-blocking-mode)
 * [Plugin Developer and Contributor Guide](#plugin-developer-and-contributor-guide)
     * [Everything is a plugin](#everything-is-a-plugin)
     * [Internal Architecture](#internal-architecture)
@@ -711,7 +713,9 @@ Now use CA flags with other
 Embed proxy.py
 ==============
 
-To start `proxy.py` in embedded mode:
+## Blocking Mode
+
+Start `proxy.py` in embedded mode by using `main` method:
 
 ```
 from proxy.main import main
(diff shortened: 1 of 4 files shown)


abhinavsingh:
@bhysyq Can you give this PR a try and see if it works well for you.  I added some instructions for running `proxy.py` in non-blocking mode, see https://github.com/abhinavsingh/proxy.py/blob/3c66d520cb9d2e0abe0a86e6264237fb1d23c668/README.md#non-blocking-mode  Thank you

bhysyq:
Hi @abhinavsingh ,

Thanks for support this feature. I tested it today, and I see the shutdown function was called in logs:
```
2019-11-06 20:12:15,058 [INFO] Shutting down 1 workers
2019-11-06 20:12:15,916 [DEBUG] Shutdown of global event dispatcher thread 6408 successful
```
Unfortunately, the port was still open, and I found it hangs at join() function in acceptor class: https://github.com/abhinavsingh/proxy.py/blob/9737a3886074e17bbee3b0c51fb6d1239f6b4af8/proxy/core/acceptor.py#L102
I also can't find `Acceptors shutdown` in my logs:
```
logger.debug('Acceptors shutdown')
```
It looks like we need to find a way to terminate run loop in acceptor.

abhinavsingh:
@bhysyq Thanks for trying out. The hanging behavior is possibly a symptom of non-daemon threads, code is essentially waiting for all of them to finish processing before shutdown.

If you have active traffic going to `proxy.py`, it can give you similar hang experience, but are you also experiencing the same hang behavior with no traffic sent to `proxy.py`?

bhysyq:
Hi  @abhinavsingh ,

I just did another test: shutting down workers, and keep sending traffic to proxy, but it seems the proxy can still handle all of them. I took close into the code, and I found the acceptor is still looping at here:
https://github.com/abhinavsingh/proxy.py/blob/3c66d520cb9d2e0abe0a86e6264237fb1d23c668/proxy/core/acceptor.py#L225
I checked the value in `self.running`, it is True, and it can not jump out of loop until exception happens (e.g. KeyboardInterrupt),

bhysyq:
Hi @abhinavsingh ,

I got the same hang behavior with no traffic sent to `proxy.py`. 
The other test I did is: shutting down workers, and stop sending traffic to proxy. I have been waiting for 20 mins, but "Acceptors shutdown" log did not show up. After 20 mins, I send traffic to proxy, and I got response, so I believe it was not shut down properly.

The environment I tested is Windows. I will try Linux later.

bhysyq:
Same in Liunx, I waited no traffic for 2 mins, here are the logs from test:
```
...... (shutdown workers)

2019-11-07 07:07:09,816 [INFO] Shutting down 1 workers
2019-11-07 07:07:10,536 [DEBUG] Shutdown of global event dispatcher thread 14055                                          5974141696 successful
2019-11-07 07:07:10,765 [DEBUG] Client buffer is empty and maximum inactivity ha                                          s reached between client and server connection, tearing down...

..... (wait until all connection are closed)

2019-11-07 07:07:49,997 [DEBUG] Closing client connection <socket.socket fd=38,                                           family=AddressFamily.AF_INET, type=2049, proto=0, laddr=('192.168.1.223', 8080),                                           raddr=('192.168.1.69', 49404)> at address ('192.168.1.69', 49404) with pending                                           client buffer size 0 bytes
2019-11-07 07:07:49,999 [DEBUG] Client connection shutdown successful
2019-11-07 07:07:50,006 [DEBUG] Client connection closed





(stop sending traffic ...)
(wait for 2 mins ...  "Acceptors shutdown" log did not show up)

(start sending traffic again)

2019-11-07 07:09:27,122 [DEBUG] Handling connection <socket.socket fd=30, family                                          =AddressFamily.AF_INET, type=2049, proto=0, laddr=('192.168.1.223', 8080), raddr                                          =('192.168.1.69', 49490)>
2019-11-07 07:09:27,128 [DEBUG] Client is ready for reads, reading
```

abhinavsingh:
Thanks a ton @bhysyq for all the analysis.  Makes sense `running` flag is still True here.  I'll update this thread once I have pushed out changes.  Ideally, `acceptor_pool.shutdown` must also toggle the `running` flag.

abhinavsingh:
@bhysyq Give it a try now, should work.

Note: It can take up-to N seconds for `proxy.py` to shutdown without interrupts, due to the way multi-accept works.  If you don't need N workers, simply use `--num-workers 1` for quick turn up and shutdown of `proxy.py`.  Let me know.  Thank you.

bhysyq:
Hi @abhinavsingh , it works as we expected now. I verified on my machine. Thanks a lot!  🥇

PR #167, opened 2019-11-12
Add proxy.main.TestCase for unit testing Python application with proxy.py

### LINKED ISSUE #162: Support for embedding Proxy.py within Python tests
I'm looking for a way to start a proxy programmatically inside test suites, but I could not find something related in the documentation, and proxy.py's own test suite or main entry points do not seem to have something like this in mind. If it can be easily used programmatically, it would be nice to document it.

abhinavsingh:
1. See [test_embed.py](https://github.com/abhinavsingh/proxy.py/blob/test-setup-teardown/tests/test_embed.py) for example usage.
2. Doc section for [Unit testing with proxy.py](https://github.com/abhinavsingh/proxy.py/tree/test-setup-teardown#unit-testing-with-proxypy)

### DIFF
--- a/README.md
+++ b/README.md
@@ -57,6 +57,10 @@ Table of Contents
 * [Embed proxy.py](#embed-proxypy)
     * [Blocking Mode](#blocking-mode)
     * [Non-blocking Mode](#non-blocking-mode)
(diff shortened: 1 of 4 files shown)


abhinavsingh:
@deeplook jFYI

PR #179, opened 2019-11-15
Move plugin_examples/ as proxy.plugin and update readme

abhinavsingh:


abhinavsingh:
@deeplook jFYI.  I think your expectation was fair :)  This PR will now bundle plugin examples with `pip install proxy.py`

Issue #251, opened 2020-01-06
-set_serial race condition can result in SEC_ERROR_REUSED_ISSUER_AND_SERIAL


JohnnySheppard:
**Describe the bug**
I'm trying out the TLS Interception. I've created my cert and keys using the makefile, and run the proxy pointing to the certificate and keys using their full path. Some sites aren't loading properly and in the proxy.py output, I'm seeing this error a lot: 

`on_request_complete:278 - OSError when wrapping client`

I'm looking at the TLS interception because I'm trying to write a plugin that will block certain url's based on their path, not just their domain - but I don't seem to be able to see the path component when the site is using TLS - and even when I've set the proxy up for TLS interception - I still can only see the domain part of the address in the proxy server. I'm assuming it's because something's not right (probably in the way I've set it up) or has something to do with that error.

It's very possible I've done something wrong in my set up, but I feel like I've followed the instructions.

**To Reproduce**
1. Run make ca-certificates
1. Run `proxy --hostname 0.0.0.0 --ca-key-file /home/pi/proxy.py/ca-key.pem --ca-cert-file /home/pi/proxy.py/ca-cert.pem --ca-signing-key-file /home/pi/proxy.py/ca-signing-key.pem `
2. Load https://www.sky.com in firefox on windows 10 setup to point to the proxy server manually.

**Expected behavior**
I would expect the page to load fully and the error not to be listed. In my plugin that I've tried - but not listed in this example (and which may also just not be working), I would expect to be able to see the path component listed as something other than they url when I dump `request.url`).

**Version information**
 - OS: Windos 10
 - Browser Firefox
 - Device: PC
 - proxy.py Version 2.0.0 running on a Raspberry Pi 3B+ with Python 3.7.3

**Additional Information**
I'm seeing the OSError with and without my plugin. My plugin is in a very rough trying to work out how to see the path component stage at the moment, and is based on the **FilterByUpstreamHostPlugin**. I don't think it is the problem but I'll list it here for completeness:
```python
#!/bin/python3

# -*- coding: utf-8 -*-
"""
    proxy.py
    ~~~~~~~~
    ⚡⚡⚡ Fast, Lightweight, Pluggable, TLS interception capable proxy server focused on
    Network monitoring, controls & Application development, testing, debugging.

    :copyright: (c) 2013-present by Abhinav Singh and contributors.
    :license: BSD, see LICENSE for more details.
"""
from typing import Optional


from proxy.http.exception import HttpRequestRejected
[...]

PR #258, opened 2020-01-07
Use UUID as request ID instead of a string. Reuse UUID to fix OSError when generating certificates

Benouare:
Hi, 
Dont really know what are the differences between -set_serial and -CAcreateserial... How they are working or the exepted behavior. 
But, regarding my test on my environnement (Mac OS 10.13.6 Python 3.8.0) it solved my issue.


abhinavsingh:
Thank you for the PR.

> The first time you use your CA to sign a certificate you can use the -CAcreateserial option. This option will create a file (ca.srl) containing a serial number. You are probably going to create more certificate, and the next time you will have to do that use the -CAserial option (and no more -CAcreateserial) followed with the name of the file containing your serial number. This file will be incremented each time you sign a new certificate. This serial number will be readable using a browser (once the certificate is imported to a pkcs12 format). And we can have an idea of the number of certificate created by a CA.

So I think `-CAcreateserial` or `-set_serial` shouldn't make a difference.  `-CAcreateserial` will simply result in a `ca.crl` file creation.

I looked at the code where `-set_serial` invocation is happening, I am wondering does changing from `time.time()` to a randomly generated number also helps fix this?  In worst case, 2 parallel requests coming to `proxy.py` can end up getting same serial number by using `time.time()`.  Wdyt? Plmk.

abhinavsingh:
.... and after looking at your investigation here https://github.com/abhinavsingh/proxy.py/issues/251#issuecomment-571527011 and https://github.com/abhinavsingh/proxy.py/issues/251#issuecomment-571534049 , I have a strong hunch that this is exactly what probably is happening.

E.g. sky.com was loaded in browser but it is trying to make say 200 requests in background.  There is a high probability that `-set_serial time.time()` will end up colliding resulting into the SSL error we ran into.

abhinavsingh:
Thank you for investigation and concluding that `-set_serial` is the problem.  I really appreciate your effort.  It's a bug and I have marked all PR / Issues accordingly.

Thinking further about it, larger problem I see is that we can run into race conditions during certificate generations, resulting into random request failures.  At the time of adding TLS interception support, I didn't verify / tested this scenario.

So how can it happen?

1. Previously in https://github.com/abhinavsingh/proxy.py/pull/258#issuecomment-571699024 I highlighted how a high throughput to `proxy.py` can result in a race condition where 2 certificates end-up with same serial ID.  Reason being usage of `time.time()` for serial ID.

2. One option can be to use `self.uid` which is a unique ID assigned to each client request by `proxy.py`.  This `self.uid` is guaranteed to be unique across all requests OR as expressed earlier, we can also generate a random uid ad-hoc for setting the serial ID.

3. I'll need to further investigate, but my gut feeling is that `openssl` is taking a lock on `-CAcreateserial` file on disk to avoid race conditions.  This is exactly what we need and hence `-CAcreateserial` is a good intermediate solution.

4. I say intermediate because I still believe we must use `-set_serial` due to performance and efficiency.   Delegating file lock and serial generation to `openssl` will only slow things down, as now `openssl` will act like a global lock shared by all `proxy.py` workers.

> Note: In-case, `openssl` is not taking a lock on `crl` file then we can still run into same issue.

Benouare:
Thx for you reply and your explications. 
The UID seems to be the good solution. In my opinion, using the request uid is better than to generate a new one.
If you want, i can try to deal with it, and make an other PR or amend my commit.

abhinavsingh:
> Thx for you reply and your explications.
> The UID seems to be the good solution. In my opinion, using the request uid is better than to generate a new one.
> If you want, i can try to deal with it, and make an other PR or amend my commit.

Sure thing, please go ahead.  Meanwhile, lemme figure why GitHub workflows are not running for your PR.

Benouare:
On my dev version it works with using the uid. 

I have to recheck my changes : and change the uid from str (`uuid.uuid4().hex()`) to the main class UUID (`uuid.uuid4()`), so i have to change all the params.
It looks like we are not able to put a Str as -set_serial, it's because of that, that i need to change the self.uid type. 

I will do that maybe tonight, or tomorrow morning.

abhinavsingh:
> On my dev version it works using the uid (with little changes...).
> I have to recheck my changes before to push: I change the uid from str (`uuid.uuid4().hex()`) to the main class UUID (`uuid.uuid4()`), so i had to change all the params in some classes.
> It looks like we are not able to put a Str as param for the -set_serial, it's because of that, that i need to change the self.uid type.
> 
> I will do that maybe tonight, or tomorrow morning.

Thanks for quickly checking out on it, I suspected it but wasn't sure.  Let's not change `uid` type because at this stage `uid` is too much spread out to change.

To fix this bug, we need a unique integer that can act as certificate serial ID.  If `time.time()` is itself running into race condition, can we do something else mitigate the race condition:

> NOTE: Client threads run within context of a particular worker.  There can be up-to N workers.

So what can we append to `time.time()` to get a unique integer _(actually it's a floating point)_.  Some options from top of the head:

a) IIRC `uid` is just a hex string so we should be able to get numbers out of it.

b) Append `time.time()` value with a random number.  E.g. say by using https://docs.python.org/3/library/random.html.  For a perfect solution with random, we should ideally seed them differently per worker process.

c) Or append `time.time()` with first k integers obtained from `uid` as discussed in a) above

One or more combination of above should result in a unique identifier.  Wdyt?

abhinavsingh:
jFYI, https://github.com/abhinavsingh/proxy.py#setup-pre-commit-hook will be useful to avoid failing workflow due to small things here and there.

Benouare:
> > On my dev version it works using the uid (with little changes...).
> > I have to recheck my changes before to push: I change the uid from str (`uuid.uuid4().hex()`) to the main class UUID (`uuid.uuid4()`), so i had to change all the params in some classes.
> > It looks like we are not able to put a Str as param for the -set_serial, it's because of that, that i need to change the self.uid type.
> > I will do that maybe tonight, or tomorrow morning.
> 
> Thanks for quickly checking out on it, I suspected it but wasn't sure. Let's not change `uid` type because at this stage `uid` is too much spread out to change.
> 
> To fix this bug, we need a unique integer that can act as certificate serial ID. If `time.time()` is itself running into race condition, can we do something else to mitigate the race condition. We don't even need an auto-incrementing ID, a random unique integer / floating point will work for our use case.
> 
> > NOTE: Client threads run within context of a particular worker.  There can be up-to N worker processes.
> 
> So can we append to `time.time()` to get a unique integer _(actually it's a floating point)_. Some options from top of the head:
> 
> a) IIRC `uid` is just a hex string so we should be able to get numbers out of it.
> 
> b) Append `time.time()` value with a random number. E.g. say by using https://docs.python.org/3/library/random.html. For a perfect solution with random, we should ideally seed them differently per worker process.
> 
> c) Or append `time.time()` with first k integers obtained from `uid` as discussed in a) above
> 
> One or more combination of above should result in a unique identifier. Wdyt?

I tried the a) solution -> int of the hex. The advantage is for me that it's already generated so with that solution i think we are not consuming more resources...

abhinavsingh:
@Benouare That sounds great.  Please send a PR to address the same :)  Thank you!!!

abhinavsingh:
> @Benouare That sounds great. Please send a PR to address the same :) Thank you!!!

Just realized you actually closed the PR, will you be sending a separate one?  I gazed over your current changes and it looks good to me.  Changing UUID from str to UUID type itself works.

abhinavsingh:
This PR looks good for merge. Waiting for tests to finish. Thanks again and apologies it took time to get back to `proxy.py`.

Benouare:
Hi Abhinav,

> This PR looks good for merge. Waiting for tests to finish. Thanks again and apologies it took time to get back to `proxy.py`.

No worries for the time ;). 
When you are saying "Waiting for tests to finish", it looks like the tests are working well, there is just 
- codecov/patch
- codecov/project
- pyup.io/safety-ci 
that are not working 
Are you waiting something about me?

abhinavsingh:
Just noticed it is `codecov` and `pyup` not updating (These checks doesn't seems to work out of box for incoming PRs).

All good, thank you, excellent investigation overall here :)

PR #261, opened 2020-01-10
SAN extension required for TLS interception certificate generation

Benouare:
Hi,

When using locally proxy.py, I saw that, with FF, everything looks good, but with Brave/Chrome that was not the case.
Thx of the console of chrome, it says that there is currently two issues on our certificates : 
- that was in sha1, not enought for chrome -> it needs sha256,
- the certificate must have SAN information (subectAltName)

This PR, should solved this issue. 
Regarding the code, the main problem was to generate on the fly certs with the SAN information.
I tried billions of time & differents solutions, but I was not able to make it on the fly.
Currently, I am generating and consuming a file, before to remove it. I think, it will make a lot of I/O access, and should be powerless... 
So, I am waiting your comment to find the best solution. 

Ressources : 
https://www.fomfus.com/articles/how-to-generate-self-signed-certificates-with-san-subject-alternative-name

Edit : 
The 'solution' was to put something like that in the signing process to add the SAN on the fly, but that's not working with python (but working well in direct shell) 
```
'-extfile','<(printf "[SAN]\nsubjectAltName=DNS:CurrentTld")'
```
But that doesnt work at all

Edit 2 : the base branch is the one for the #258, hope that's not a problem for you.

### DIFF
--- a/proxy/common/pki.py
+++ b/proxy/common/pki.py
@@ -16,7 +16,7 @@ import uuid
 import subprocess
 import tempfile
 import logging
-from typing import List, Generator, Optional, Tuple
+from typing import Generator, List, Optional, Tuple
 
 from .utils import bytes_
 from .constants import COMMA
@@ -50,6 +50,7 @@ emailAddress_max		= 64
 challengePassword		= A challenge password
 challengePassword_min	= 4
 challengePassword_max	= 20'''
+SAN_KEY = "SAN"
 
 
 def remove_passphrase(
@@ -101,23 +102,22 @@ def gen_public_key(
         ]
         if has_extension:
             command.extend([
-                '-extensions', 'PROXY',
+                '-extensions', SAN_KEY,
             ])
         return run_openssl_command(command, timeout)
 
 
 def gen_csr(
         csr_path: str,
-        key_path: str,
-        password: str,
-        crt_path: str,
+        signing_key_path: str,
+        subject: str,
+        validity_in_days: int = 365,
         timeout: int = 10) -> bool:
     """Generates a CSR based upon existing certificate and key file."""
     command = [
-        'openssl', 'x509', '-x509toreq',
-        '-passin', 'pass:%s' % password,
-        '-in', crt_path, '-signkey', key_path,
-        '-out', csr_path
+        'openssl', 'req', '-new', '-sha256',
+        '-days', str(validity_in_days), '-subj', subject,
+        '-key', signing_key_path, '-out', csr_path
     ]
     return run_openssl_command(command, timeout)
 
@@ -126,10 +126,9 @@ def sign_csr(
         csr_path: str,
         crt_path: str,
         ca_key_path: str,
-        ca_key_password: str,
         ca_crt_path: str,
         serial: str,
-        alt_subj_names: Optional[List[str]] = None,
+        alt_subj_names: List[str],
         extended_key_usage: Optional[str] = None,
         validity_in_days: int = 365,
         timeout: int = 10) -> bool:
@@ -139,28 +138,42 @@ def sign_csr(
             'openssl', 'x509', '-req', '-sha256',
             '-CA', ca_crt_path,
             '-CAkey', ca_key_path,
-            '-passin', 'pass:%s' % ca_key_password,
             '-set_serial', serial,
             '-days', str(validity_in_days),
             '-extfile', extension_path,
+            '-extensions', SAN_KEY,
             '-in', csr_path,
             '-out', crt_path,
         ]
         return run_openssl_command(command, timeout)
 
 
-def get_ext_config(
-        alt_subj_names: Optional[List[str]] = None,
-        extended_key_usage: Optional[str] = None) -> bytes:
+def gen_crt(
+        crt_path: str,
+        signing_key_path: str,
+        ca_key_path: str,
+        ca_crt_path: str,
+        tld: str,
+        serial: str) -> None:
+    """For a given tld, generate an signed cert."""
+    temp_csr = os.path.join(tempfile.gettempdir(), uuid.uuid4().hex)
+    gen_csr(temp_csr, signing_key_path, f'/C=/ST=/L=/O=/OU=/CN={tld}')
+    sign_csr(temp_csr, crt_path,
+             ca_key_path, ca_crt_path, serial, [tld])
+
+
+def get_ext_config(alt_subj_names: Optional[List[str]] = None, extended_key_usage: Optional[str] = None) -> bytes:
     config = b''
     # Add SAN extension
     if alt_subj_names is not None and len(alt_subj_names) > 0:
+        config += b'[' + bytes_(SAN_KEY) + b']'
         alt_names = []
         for cname in alt_subj_names:
             alt_names.append(b'DNS:%s' % bytes_(cname))
         config += b'\nsubjectAltName=' + COMMA.join(alt_names)
     # Add extendedKeyUsage section
     if extended_key_usage is not None:
+        config += b'[PROXY]'
         config += b'\nextendedKeyUsage=' + bytes_(extended_key_usage)
     return config
 
@@ -191,7 +204,6 @@ def ssl_config(
     if (alt_subj_names is not None and len(alt_subj_names) > 0) or \
             extended_key_usage is not None:
         has_extension = True
-        config += b'\n[PROXY]'
 
     # Add custom extensions
     config += get_ext_config(alt_subj_names, extended_key_usage)
@@ -219,8 +231,7 @@ def run_openssl_command(command: List[str], timeout: int) -> bool:
 
 if __name__ == '__main__':
     available_actions = (
-        'remove_passphrase', 'gen_private_key', 'gen_public_key',
-        'gen_csr', 'sign_csr'
+        'remove_passphrase', 'gen_private_key', 'gen_public_key'
     )
 
     parser = argparse.ArgumentParser(
--- a/proxy/http/proxy/server.py
+++ b/proxy/http/proxy/server.py
@@ -9,7 +9,6 @@
     :license: BSD, see LICENSE for more details.
 """
 import threading
-import subprocess
 import os
 import ssl
 import socket
@@ -25,6 +24,7 @@ from ..codes import httpStatusCodes
 from ..parser import HttpParser, httpParserStates, httpParserTypes
 from ..methods import httpMethods
 
+from ...common import pki
 from ...common.types import HasFileno
 from ...common.constants import PROXY_AGENT_HEADER_VALUE
 from ...common.utils import build_http_response, text_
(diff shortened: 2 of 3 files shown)


abhinavsingh:
@Benouare I think we must start to use `pki.py` utilities within the interception code for generating and signing CSR requests.  See `test_pki.py` for some usage information.

While `pki.py` was updated to comply with latest SSL cert requirements, I think hardcoded `openssl` usages within interception code might still be complying with cert requirements _(specially the extensions part of cert)_.

Benouare:
> @Benouare I think we must start to use `pki.py` utilities within the interception code for generating and signing CSR requests. See `test_pki.py` for some usage information.
> 
> While `pki.py` was updated to comply with latest SSL cert requirements, I think hardcoded `openssl` usages within interception code might still be complying with cert requirements _(specially the extensions part of cert)_.

I agree, I will make the corrections.

Benouare:
So, I have made some changes. 
I push the branch but it's not the final 'release' version of it : we need to talk about it, and i would like your feedback.
I have to delete some part of the code to generate the certificates 'in the good way'. 
I tried to use the existants pki methods to generate certs, but that doesnt work, so I do that.
About the test, I add some, but there is one that is not working anymore. Dont get why... 
Let me know what you think about it.

abhinavsingh:
Sure thing, I'll take a look at it tonight and get back to you.  Thank
you!!!

abhinavsingh:
Hi @Benouare ,

I am wondering why do we need new `gen_web_csr` and `gen_web_cert` methods, doesn't existing methods in `pki.py` already suffice?  We already used them previously to update our `Makefile`.

I am thinking of putting those 4 loc that you added under `gen_web` to replace current `openssl` commands run directly during interception.  Wdyt?

Also, let's separate out your PR's.  Send out a separate one for `UUID` updates, we should be able to check that in sooner.  Then let's continue on this PR for cert generation fixes.

Benouare:
> I am wondering why do we need new `gen_web_csr` and `gen_web_cert` methods, doesn't existing methods in `pki.py` already suffice? We already used them previously to update our `Makefile`.

I did'nt succeed to use the methods in the pki, to generate certs that can be used by ff/chrome & others. It's for that reason that I have put some new methods. 

> I am thinking of putting those 4 loc that you added under `gen_web` to replace current `openssl` commands run directly during interception. Wdyt?

I don't get what you mean. The gen_web is currently used to replace the 'old' way to generate the certificate during the interception.

> Also, let's separate out your PR's. Send out a separate one for `UUID` updates, we should be able to check that in sooner. Then let's continue on this PR for cert generation fixes.

Get it.

abhinavsingh:
> > I am wondering why do we need new `gen_web_csr` and `gen_web_cert` methods, doesn't existing methods in `pki.py` already suffice? We already used them previously to update our `Makefile`.
> 
> I did'nt succeed to use the methods in the pki, to generate certs that can be used by ff/chrome & others. It's for that reason that I have put some new methods.
> 

I am trying to understand why can't we modify existing `pki.py` methods to make them compatible with FF/Chrome.  As of today, `pki.py` is not used by anyone except `Makefile`, so it is OK to make changes directly into `pki.py` gen_priv_key and gen_public_key methods.

IIUC, only different is in terms of a few flags right?

Benouare:
> I am trying to understand why can't we modify existing `pki.py` methods to make them compatible with FF/Chrome.
> As of today, `pki.py` is not used by anyone except `Makefile`, so it is OK to make changes directly into `pki.py` gen_priv_key and gen_public_key methods.
> 
> IIUC, only different is in terms of a few flags right?

proxy/http/proxy/server.py
```diff
@@ -274,8 +274,9 @@ def on_request_complete(self) -> Union[socket.socket, bool]:
                     # wrap_client also flushes client data before wrapping
                     # sending to client can raise, handle expected exceptions
                     self.wrap_client()
-                except OSError:
-                    logger.error('OSError when wrapping client')
+                except OSError as e:
+                    logger.error(
+                        'OSError when wrapping client :"{}"'.format(e.strerror))
```
Benouare: just to help a little...

proxy/common/pki.py
```diff
@@ -81,87 +56,72 @@ def gen_private_key(
...
+            signing_key_path: str,
+            key_path: str,
+            crt_path: str,
+            tld: str,
+            serial: str) -> None:
+    """For a given tld, generate an signed cert."""
+    temp_cert = os.path.join(tempfile.gettempdir(), uuid.uuid4().hex)
+    gen_web_public_key(temp_cert, signing_key_path,
+                       f'/C=/ST=/L=/O=/OU=/CN={tld}')
+    sign_csr(csr_path, key_path, crt_path, temp_cert, serial, [tld])
```
Benouare: FYI, if i add a 
```
os.remove(temp_cert)
```
The e2e test will fail : it looks like the temp_cert is already deleted, but I dont know by who. 
It looks like the OS is cleaning the temp dir automatically, so I propose to dont put the os.remove. 
I let u think about to drop it in other part of the code and let the OS manage those temp files.

abhinavsingh:
Thanks again, I am reviewing and revisiting this today.

IIUC here are the outstanding items we need to address here:

1) Chrome needs SAN extension which is not part of current signing process
2) We are unable to re-use existing `pki.py` methods

2nd is what I want to dig further today.  I think we might be able to just re-use.  As expressed earlier, I wrote `pki.py` to replace all signing stuff, but never got to it.

I'll update back with my findings.  Thank you.

Benouare:
> IIUC here are the outstanding items we need to address here:
> 
> 1. Chrome needs SAN extension which is not part of current signing process
> 2. We are unable to re-use existing `pki.py` methods

1. SAN and SHA 256
2. More or less yes. I try many things to deal with be able to use the existing method. But didnt succeed. 

Maybe you can try to do something with the current PR (or something else...)

abhinavsingh:
> > IIUC here are the outstanding items we need to address here:
> > 
> > 1. Chrome needs SAN extension which is not part of current signing process
> > 2. We are unable to re-use existing `pki.py` methods
> 
> 1. SAN and SHA 256
> 2. More or less yes. I tried many things to be able to use the existing method. But didnt succeed.
> 
> Maybe you can try to do something with the current PR (or something else...)

Lemme dig into 2nd item in detail. In case we could re-use, I'll update back with my findings 👍

proxy/common/pki.py
```diff
@@ -81,87 +56,72 @@ def gen_private_key(
     return run_openssl_command(command, timeout)
 
 
-def gen_public_key(
+def gen_web_public_key(
         public_key_path: str,
         private_key_path: str,
-        private_key_password: str,
         subject: str,
-        alt_subj_names: Optional[List[str]] = None,
```
abhinavsingh: I don't think we can do this. Also, I think have a different notion of how TLS interception works (I will post about it in separate review comment)

Having alt_subj_names in certificates is a requirement per Mac OS new security policies.  See this https://support.apple.com/en-us/HT210176

proxy/http/proxy/server.py
abhinavsingh: Sgtm

proxy/http/proxy/server.py
```diff
@@ -360,20 +361,8 @@ def generate_upstream_certificate(
         with self.lock:
             if not os.path.isfile(cert_file_path):
                 logger.debug('Generating certificates %s', cert_file_path)
-                # TODO: Parse subject from certificate
-                # Currently we only set CN= field for generated certificates.
```
abhinavsingh: Do preserve TODOs.  Someday it will be worth addressing these.

proxy/http/proxy/server.py
```diff
@@ -360,20 +361,8 @@ def generate_upstream_certificate(
...
-                    stderr=subprocess.PIPE)
-                sign_cert = subprocess.Popen(
-                    ['openssl', 'x509', '-req', '-days', '365', '-CA', self.flags.ca_cert_file, '-CAkey',
-                     self.flags.ca_key_file, '-set_serial', str(self.uid.int), '-out', cert_file_path],
-                    stdin=gen_cert.stdout,
-                    stderr=subprocess.PIPE)
-                # TODO: Ensure sign_cert success.
-                sign_cert.communicate(timeout=10)
+                pki.gen_web(cert_file_path, self.flags.ca_signing_key_file, self.flags.ca_key_file,
+                            self.flags.ca_cert_file, text_(self.request.host), str(self.uid.int))
```
abhinavsingh: I kind of fail to understand the motive behind all the changes.

a) I think they won't work cross platform e.g. will break on MacOS
b) IIUC, all we wanted was `SAN` extension in the signing step.  So you could have only modified `sign_csr` to incorporate `SAN` extension when called from `server.py`

proxy/common/pki.py
Benouare: Thx for the link. 
The strange part is that, FF doesnt care of those "rules".

proxy/http/proxy/server.py
Benouare: We need SAN and SHA 256. As your link say : 
> SHA-1 signed certificates are no longer trusted for TLS. 

If i remember well, we must have a base cert/key/ca in sha256, to sign in sha256. It's for this reason that I change a big part of the implementation.
I will take a look on it today.

proxy/common/pki.py
Benouare: It looks like it's a mandatory field because of chrome, and not only because of OSX => https://support.google.com/chrome/a/answer/7391219?hl=en

proxy/http/proxy/server.py
Benouare: Done.

proxy/common/pki.py
abhinavsingh: Exactly. SAN is just needed during certificate signing, so my suggestion is to only change that part of the code. So there are 2 steps happening during interception:

1) Generate a CSR for domain
2) Sign CSR with CA

To generate a CSR we also need domain private key. This is what current code does, it generate an interim key (never saved on disk), pipes the output into CSR signing command.

We'll have to add a similar method where CSR can be generated without having to persist private key `(e.g. gen_priv_key, then remove_passphrase, then gen_csr)`.  You can also remove the need of `remove_passphrase` by not encrypting the private key in first place (i.e. don't use a password while generating one, this will require changes to `gen_priv_key` flags as it by default only generated encrypted private key).  Note that, CSR generation is a one time step and valid till 365 days (current default sign validity).

Also, utilities in `pki.py` are not in anyway tied with its user.  Idea is to simply have all PKI utilities at a central place for re-use.  So, I think terms like "web_cert" shouldn't land in `pki.py`, just continue using standard terminology like private / public keys and signed certificates (for method names).

Hence we must also preserve `DEFAULT_CONFIG` for use when needed. We can make ways to take input custom config `(e.g. SAN during signing)`, overriding default config.

I hope this background helps you refactor better :)

abhinavsingh:
@Benouare Thanks for your patience.  These have been tough times specially with all the travel involved.  Good news, I am back at a stable place and will soon jump into pending PR.  Really appreciate it.  I hope you are safe and sound in these hard times.  Thank you!!!

abhinavsingh:
This should be resolved with #362 .  Apologies for having to close this.  I think approach in #362 is simpler (a single method added to `server.py`) and if it works let's persist with it.  Do check it out, thank you for all the help and investigation done on this subject.

PR #269, opened 2020-01-21
Fix for urllib with python 3.7+ & 3.8+

Benouare:


Benouare:
I just thought that i didnt check the retrocompatibility... I think, we will not be able to use proxy.py with '3.7-' and others... 
Let's see what github tests will say....

Benouare:
Is that normal that, the python **3.6** process is using **3.8.1**? 
Capture d’écran 2020-01-21 à 11 48 46

abhinavsingh:
> Is that normal that, the python **3.6** process is using **3.8.1**?

Yes I have seen this with GitHub Actions.  They don't seem to respect Python versions specified.  We can also use bundled [tox.in](https://github.com/abhinavsingh/proxy.py/blob/develop/tox.ini) to test verify `proxy.py` across all Python versions locally.

I am also reviewing PR now :) Thank you!!!

abhinavsingh:
I think we must get this out asap.  I'll try to contribute into your PR tonight, so that we can ship this asap.

1) Please remove UUID changes from your PR.  Let's handle that separately.
2) In this PR only change could be appending `https://` for `proxy.py` connect request lines.

`proxy.py` download numbers have fallen from 25k to 18k per month.  Of-course people are not even able to use it right now.

Benouare:
I have just checked, this branch is not working at all.

PR #271, opened 2020-01-22
Fix for urllib with python 3.7+ & 3.8+

Benouare:
Related to : https://github.com/abhinavsingh/proxy.py/pull/269

### DIFF
--- a/requirements-testing.txt
+++ b/requirements-testing.txt
@@ -5,7 +5,6 @@ pytest==5.3.2
 pytest-cov==2.8.1
 autopep8==1.4.4
 mypy==0.761
-py-spy==0.3.0
 codecov==2.0.15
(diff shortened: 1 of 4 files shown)


requirements-testing.txt
```diff
@@ -5,7 +5,6 @@ pytest==5.3.2
 pytest-cov==2.8.1
 autopep8==1.4.4
 mypy==0.761
-py-spy==0.3.0
```
Benouare: FYI, if we don't drop this, the pip install -r will not work : 
`Could not find a version that satisfies the requirement py-spy==0.3.0`

abhinavsingh:
This was fixed in https://github.com/abhinavsingh/proxy.py/pull/277 , closing now.

PR #272, opened 2020-01-22
Add WebServerPlugin automatically at startup if needed

### LINKED ISSUE #256: Warn user about missing --enable-web-server flag when a web server plugin is loaded
Users have reached out with questions like

Screen Shot 2020-01-06 at 9 43 26 PM

Warn (or abort?) if user has loaded a http web server plugin but haven't specified `--enable-web-server` flag.

**Describe alternatives you've considered**
Alternately, `--enable-web-server` flag can be turned on implicitly if a web server plugin is detected.

Benouare:
Fixes #256


proxy/common/flags.py
```diff
@@ -9,31 +9,38 @@
...
-from .constants import DEFAULT_LOG_LEVEL, DEFAULT_LOG_FILE, DEFAULT_LOG_FORMAT, DEFAULT_BACKLOG, DEFAULT_BASIC_AUTH
-from .constants import DEFAULT_TIMEOUT, DEFAULT_DEVTOOLS_WS_PATH, DEFAULT_DISABLE_HTTP_PROXY, DEFAULT_DISABLE_HEADERS
-from .constants import DEFAULT_ENABLE_STATIC_SERVER, DEFAULT_ENABLE_EVENTS, DEFAULT_ENABLE_DEVTOOLS
-from .constants import DEFAULT_ENABLE_WEB_SERVER, DEFAULT_THREADLESS, DEFAULT_CERT_FILE, DEFAULT_KEY_FILE
-from .constants import DEFAULT_CA_CERT_DIR, DEFAULT_CA_CERT_FILE, DEFAULT_CA_KEY_FILE, DEFAULT_CA_SIGNING_KEY_FILE
-from .constants import DEFAULT_PAC_FILE_URL_PATH, DEFAULT_PAC_FILE, DEFAULT_PLUGINS, DEFAULT_PID_FILE, DEFAULT_PORT
-from .constants import DEFAULT_NUM_WORKERS, DEFAULT_VERSION, DEFAULT_OPEN_FILE_LIMIT, DEFAULT_IPV6_HOSTNAME
-from .constants import DEFAULT_SERVER_RECVBUF_SIZE, DEFAULT_CLIENT_RECVBUF_SIZE, DEFAULT_STATIC_SERVER_DIR
-from .constants import DEFAULT_ENABLE_DASHBOARD, COMMA, DOT
+from .constants import (COMMA, DEFAULT_BACKLOG, DEFAULT_BASIC_AUTH,
```
abhinavsingh: Can you configure your editor to not modify a line of code that wasn't changed?  Otherwise it becomes harder to review between a) cosmetic changes b) actual logic changes I generally don't like to fiddle with import orders (may be it's just your editor).

proxy/common/flags.py
```diff
@@ -523,6 +530,13 @@ def load_plugins(plugins: bytes) -> Dict[bytes, List[type]]:
                 # HttpWebServerRouteHandler route decorator adds a special
                 # staticmethod to return decorated function name
                 klass.__name__ if klass.__name__ != 'HttpWebServerRouteHandler' else klass.name())
+        # if a plugin need the HttpWebServerPlugin to be loaded, and if it's missing, we force the load of this plugin
```
abhinavsingh: I am also not sure if we must add this logic inside `load_plugins` which is actually agnostic of any logic imposed by `proxy.py`.  All it knows is, load a list of plugins passed to it.  So adding a business logic like `enable plugin if / else` here won't be a cleaner solution.  Let's enforce HttpWebServerPlugin at a high level.

I was thinking of adding necessary checks within Flags module, but before `load_plugin` gets called.  Idea is that we should be able to detect before calling load_plugins, whether web server plugin should be loaded or not.

Benouare: > I was thinking of adding necessary checks within Flags module, but before load_plugin gets called. Idea is that we should be able to detect before calling load_plugins, whether web server plugin should be loaded or not.

I agree with you. My first idea was something like that, but my main problem was to get the "main" class of each plugin.  It makes us able to know if we must add the "--enable-web-server"/"HttpWebServerPlugin" or not. 
But, to do that, it seems that we are obliged to do all the process  as you currently do with
```
klass = getattr(
                importlib.import_module(
                    module_name.replace(
                        os.path.sep, text_(DOT))),
                klass_name)
```
to get the main class for each plugin.

If we check before the call of load_plugins, the behavior of this check will be more all less the same because, as I said before, it looks like we need to get the `klass` to know the main class of the plugin.

proxy/common/flags.py
Benouare: I have configure my editor (atom) with default python conf. I will change it that it will doesnt change the import orders.

abhinavsingh:
@Benouare Hey, closing this as this seems outdated for now.  I remember we had a hanging discussion around `load_plugins` and `import_module` usage.  Let's revisit later again.  We need to revamp the flags module completely to allow custom flags per individual plugins instead of adding everything under same global namespace.  Thank you!!!

PR #303, opened 2020-03-08
ReverseProxyPlugin sporadically fails to process requests

atsakiridis:
This seems to be fixing the issue, even though still a draft. Need to iterate to make more modular and potentially more efficient


abhinavsingh:
@atsakiridis Hey closing this now.  We still need to revisit this and provide plugins a way to register custom sockets for eventing.  A couple of more pending items before we jump to that though e.g. revamping the flags module to allow custom flags per plugin instead of adding all the flags under the same global namespace.  Thank you!!!

PR #342, opened 2020-05-20
Add ModifyChunkResponsePlugin example. Fixes #332.

### LINKED ISSUE #332: [MITM] Changing response body
First off, thanks for a great library! 

I've successfully implemented a basic MITM proxy with TLS. However, I'm trying to work out how to efficiently parse the response from the upstream server, inspect the contents and change the data being sent downstream to the original requester.

For example, the requester is attempting to download a JSON file as octet/binary; I want to catch the response and add fields to the JSON being returned. As soon as I make changes to the JSON data, the response to the client hangs or the JSON data is truncated. I suspect I'm doing something wrong with the ``HttpParser`` as the value of ``parser.body`` is always ``None``.

What I'm seeing though makes sense as the length of the body I'm returning is longer (rather, different) to the value of the Content-Length header; I need to change that value, but the headers and the JSON data are being returned in two different chunks.

Is there another way to create a plugin which achieve this?

abhinavsingh:
@anthonykgross @grvhi PTAL, one can expand on this to achieve the necessary.

## Test It

Start `proxy.py`

`python -m proxy --plugin proxy.plugin.ModifyChunkResponsePlugin --log-level d`

Make a request which sends chunked responses

`curl -v -x localhost:8899 http://httpbin.org/stream/10`

## Modify Response

Instead of accumulating and dispatching response to client, one may send modified response.  Example, replace `self.client.queue(memoryview(b''.join(self.chunks)))` with `self.client.queue(memoryview(b'My Funky Response'))`

## Limitations

As you may observe, after receiving all chunks from upstream, `ModifyChunkResponsePlugin` has to wait until upstream server connection times out, since `ModifyChunkResponsePlugin` doesn't have a way to identify whether more chunks are expected.

This can be solved by adding a special marking within `HttpParser` which can annotate the responses as "last chunk received".  Using this information, `ModifyChunkResponsePlugin` can then immediately dispatch the response to client.


abhinavsingh:
@anthonykgross @grvhi  This example is now officially in the repo.  See https://github.com/abhinavsingh/proxy.py/pull/387 and related docs.  Thank you.

PR #370, opened 2020-06-13
Copy certificate subject from upstream

abhinavsingh:
Additionally, use `ssl.PROTOCOL_TLSv1_2` for client wrap and allow `TLSv1_1` for server wrap


abhinavsingh:
Will result in following certificate when inspected from browser

Screen Shot 2020-06-13 at 9 30 07 PM

PR #420, opened 2020-08-11
Allow to use types when embeding Proxy

pasccom:
When embedding `Proxy` in another application (e.g. automatic tests), it may be simpler to directly give the `type` of the plugins to load via the `plugins` keyword argument. 
This requires to convert the `plugins` keyword argument into a `list` of `bytes` or `type` and modify `Flags.load_plugin`. I corrected the test so that they comply to the new type of the `plugin` keyword argument.
I also included unit tests for `Flags.load_plugin` as well as a correction ensuring plugins are only loaded once (see the three last unit tests).


abhinavsingh:
This looks promising.  I will be away from work laptop today & tomorrow.  Will soon get to it soon thereafter.  Thank you!!!

abhinavsingh:
@pasccom At some point we must also document it for discovery.  A link to tests under `embed` section of `README.md` will help.  It's ok to add this in a follow up PR if you wish.

pasccom:
> At some point we must also document it for discovery. A link to tests under `embed` section of `README.md` will help. It's ok to add this in a follow up PR if you wish.

I should have thought about this... I had to read the code to understand how to load plugins in embedded mode. 
I think we can give some examples directly in the `README.md`. It would be simpler, wouldn't it? I will do this right now.

abhinavsingh:
Totally, examples will provide best context.  We can add this under Embed
Proxy.py section somewhere.

Thank you

PR #422, opened 2020-08-14
Documentation for plugin loading in embedded mode

pasccom:
Here is the documentation for #420 

I stuck with the 2-spaces indentation which was used in this section.
Fell free to correct any bad English: I am not a native speaker.


abhinavsingh:
Thanks you, looks perfect to me.

Issue #319, opened 2020-03-24
[CacheResponsesPlugin] Support serving out of cache


abhinavsingh:
Currently [cache_responses.py](https://github.com/abhinavsingh/proxy.py#cacheresponsesplugin) plugin only caches but doesn't serves out of cached data.

Per @trianta2 request [here](https://github.com/abhinavsingh/proxy.py/issues/301#issuecomment-596126525) it will be a good idea to support this in future releases.

However, for a production grade usage, this feature will require significant work.  If someone is interested in taking a stab at this one, please feel free to reach out.  Happy to discuss it further.

ja8zyjits:
On the outset it appears as a straight forward approach to map the cached content to the saved url.

Is that the only thing that needs to be done? Can you elaborate the production grade requirement?

abhinavsingh:
Cache invalidation is the key for production grade caching solutions.  E.g. Upstream servers will serve a URL with some cache control/expiry headers.  An ideal cache server must invalidate cache and fetch/serve fresh response after cache expiry.  Over the course of HTML specification upgrades, there are just too many cache headers to take care of e.g. `vary`, `etag`, ... etc

ja8zyjits:
That would mean replicating the entire HTTP cache protocol with active support for specification changes over every new releases?

abhinavsingh:
Not necessarily.  We can start adding subset of features (enabled explicitly via flags) as needed by the community.

Probably also a good idea to add references to section(s) of RFC that cache plugin implements in the documentation.

Happy to discuss if you have any particular scenario in mind.

ja8zyjits:
If we start small, from where must we begin?
By RFC did you mean this https://tools.ietf.org/html/rfc7234 ?

abhinavsingh:
@ja8zyjits Yep that's the RFC we can follow.  Ideally, we must pick a use case that your application currently needs.  If you would like to start working towards it generically, I'll recommend following steps:

1) Start reading the RFC as much as you can :) Especially go through section 2, 3, 4, 5.  Section 2 and 3 are rather short introductions.  Section 4 describes various scenarios and strategies.  Section 5 provide details about various cache headers and how to handle them.

2) For implementation pick a use case from section 5.  E.g. handling header fields like `Age`, `Cache-Control`, `max-age`, `no-store`, `no-cache`, `must-revalidate`, ... etc

Let me know where you would like to start.  Thank you!!!

ja8zyjits:
[...]

PR #425, opened 2020-08-18
[CacheResponsesPlugin] Serving out of cache

pasccom:
Here is a basic implementation of a cache plugin which supports serving responses out of cache. This should be the base of the solution to #319. Tests are attached for caching and serving HTTP and HTTPS requests.

The implementation is done as follows. A list file (`list.txt`) has been added which list all completed requests (method, host, path and SHA512 hash of body). When a request arrives, the plugin searches the list for this request. If it is not stored, create the entry and fetch response from server, otherwise queue the cached response to the client (and do not connect to the server).

It does not support interception of SSL handshake, which means it needs to connect to the server even though the request is cached. I think this requires a modification of the `proxy.http.proxy.server` class, so that it can perform an SSL handshake with the client without performing one with an upstream server.

### DIFF
--- a/proxy/plugin/cache/base.py
+++ b/proxy/plugin/cache/base.py
@@ -11,9 +11,13 @@
 import logging
 from typing import Optional, Any
 
-from ...http.parser import HttpParser
-from ...http.proxy import HttpProxyBasePlugin
 from .store.base import CacheStore
+from ...http.parser import HttpParser, httpParserTypes
+from ...http.proxy import HttpProxyBasePlugin
+from ...http.codes import httpStatusCodes
+from ...common.constants import PROXY_AGENT_HEADER_VALUE
+from ...common.utils import text_
+from ...common.utils import build_http_response
 
 logger = logging.getLogger(__name__)
 
@@ -40,16 +44,60 @@ class BaseCacheResponsesPlugin(HttpProxyBasePlugin):
     def before_upstream_connection(
             self, request: HttpParser) -> Optional[HttpParser]:
         assert self.store
+        logger.info("Upstream connexion %s:%d %s" %
+                    (text_(request.host), request.port if request.port else 0, text_(request.path)))
+
+        if request.port == 443:
+            return request
+
         try:
-            self.store.open(request)
+            if self.store.is_cached(request):
+                return None
         except Exception as e:
-            logger.info('Caching disabled due to exception message %s', str(e))
+            logger.info('Caching disabled due to exception message: %s', str(e))
+
         return request
 
     def handle_client_request(
             self, request: HttpParser) -> Optional[HttpParser]:
         assert self.store
-        return self.store.cache_request(request)
+        logger.info("Client request %s:%d %s" %
+                    (text_(request.host), request.port if request.port else 0, text_(request.path)))
+
+        if request.port == 443:
+            return request
+
+        try:
+            msg = self.store.cache_request(request)
+            if (msg.type == httpParserTypes.REQUEST_PARSER):
+                return msg
+            elif (msg.type == httpParserTypes.RESPONSE_PARSER):
+                self.client.queue(memoryview(build_http_response(
+                    int(msg.code) if msg.code is not None else 0,
+                    reason=msg.reason,
+                    headers={k: v for k, v in msg.headers.values()},
+                    body=msg.body
+                )))
+                return None
+            else:
+                raise ValueError('Bad HTTPParser type: %s' % msg.type)
+        except Exception as e:
+            logger.info('Caching disabled due to exception message: %s', str(e))
+
+        try:
+            if self.store.is_cached(request):
+                self.client.queue(memoryview(build_http_response(
+                    httpStatusCodes.INTERNAL_SERVER_ERROR,
+                    reason=b'Internal server error',
+                    headers={
+                        b'Server': PROXY_AGENT_HEADER_VALUE,
+                        b'Connection': b'close',
+                    }
+                )))
+        except Exception as e:
+            logger.info('Caching disabled due to exception message: %s', str(e))
+
+        return request
 
     def handle_upstream_chunk(self, chunk: memoryview) -> memoryview:
         assert self.store
--- a/proxy/plugin/cache/base.py
+++ b/proxy/plugin/cache/base.py
@@ -9,6 +9,7 @@
     :license: BSD, see LICENSE for more details.
 """
 import logging
+import multiprocessing
 from typing import Optional, Any
 
 from .store.base import CacheStore
@@ -31,18 +32,37 @@ class BaseCacheResponsesPlugin(HttpProxyBasePlugin):
     Different storage backends can be used per request if required.
     """
 
+    class EnabledDescriptor:
+        def __init__(self, enabled: bool = False) -> None:
+            self.enabled = multiprocessing.Event()
+            if enabled:
+                self.enabled.set()
+
+        def __get__(self, obj: Optional[object], owner: type) -> Any:
+            if obj is None:
+                return self.enabled
+            return None
+
+    # Dynamically enable / disable cache
+    enabled = EnabledDescriptor(True)
+
     def __init__(
             self,
             *args: Any,
             **kwargs: Any) -> None:
         super().__init__(*args, **kwargs)
         self.store: Optional[CacheStore] = None
+        self.__enabled: Optional[bool] = None
 
     def set_store(self, store: CacheStore) -> None:
         self.store = store
 
     def before_upstream_connection(
             self, request: HttpParser) -> Optional[HttpParser]:
+        self.__enabled = self.__class__.enabled.is_set()
+        if not self.__enabled:
+            return request
+
         assert self.store
         logger.info("Upstream connexion %s:%d %s" %
                     (text_(request.host), request.port if request.port else 0, text_(request.path)))
@@ -60,6 +80,10 @@ class BaseCacheResponsesPlugin(HttpProxyBasePlugin):
 
     def handle_client_request(
             self, request: HttpParser) -> Optional[HttpParser]:
+        assert self.__enabled is not None
+        if not self.__enabled:
+            return request
+
         assert self.store
         logger.info("Client request %s:%d %s" %
                     (text_(request.host), request.port if request.port else 0, text_(request.path)))
@@ -100,9 +124,17 @@ class BaseCacheResponsesPlugin(HttpProxyBasePlugin):
         return request
 
     def handle_upstream_chunk(self, chunk: memoryview) -> memoryview:
+        assert self.__enabled is not None
+        if not self.__enabled:
+            return chunk
+
         assert self.store
         return self.store.cache_response_chunk(chunk)
 
     def on_upstream_connection_close(self) -> None:
+        assert self.__enabled is not None
+        if not self.__enabled:
+            return
+
         assert self.store
         self.store.close()
--- a/proxy/plugin/cache/store/disk.py
+++ b/proxy/plugin/cache/store/disk.py
@@ -10,6 +10,7 @@
 """
 import logging
 import os
+import multiprocessing
 from typing import Optional, BinaryIO
 from hashlib import sha512
 from uuid import UUID
@@ -23,6 +24,7 @@ logger = logging.getLogger(__name__)
 
 
 class OnDiskCacheStore(CacheStore):
+    lock = multiprocessing.Lock()
 
     def __init__(self, uid: UUID, cache_dir: str) -> None:
         super().__init__(uid)
@@ -43,13 +45,14 @@ class OnDiskCacheStore(CacheStore):
     def cache_request(self, request: HttpParser) -> Optional[HttpParser]:
         body_hash = sha512(request.body).hexdigest() if request.body else 'None'
         cache_file_name = '%s-%s.txt' % (text_(request.host), self.uid.hex)
-        self.cache_list.write('%s %s %s %s %s\n' % (
-            request.method.decode() if request.method else 'None',
-            request.host.decode() if request.host else 'None',
-            request.path.decode() if request.path else 'None',
-            body_hash,
-            cache_file_name
-        ))
+        with self.lock:
+            self.cache_list.write('%s %s %s %s %s\n' % (
+                request.method.decode() if request.method else 'None',
+                request.host.decode() if request.host else 'None',
+                request.path.decode() if request.path else 'None',
+                body_hash,
+                cache_file_name
+            ))
 
         if self.cache_file:
             self.cache_file.close()
--- a/proxy/plugin/cache/store/disk.py
+++ b/proxy/plugin/cache/store/disk.py
@@ -30,12 +30,14 @@ class OnDiskCacheStore(CacheStore):
         self.cache_file: Optional[BinaryIO] = None
 
     def open(self, request: HttpParser) -> None:
-        self.cache_file_path = os.path.join(
-            self.cache_dir,
-            '%s-%s.txt' % (text_(request.host), self.uid.hex))
-        self.cache_file = open(self.cache_file_path, "wb")
+        pass
 
     def cache_request(self, request: HttpParser) -> Optional[HttpParser]:
+        self.cache_file_path = os.path.join(self.cache_dir, '%s-%s.txt' % (text_(request.host), self.uid.hex))
+
+        if self.cache_file:
+            self.cache_file.close()
+        self.cache_file = open(self.cache_file_path, "ab")
         return request
 
     def cache_response_chunk(self, chunk: memoryview) -> memoryview:
--- a/proxy/plugin/cache/store/disk.py
+++ b/proxy/plugin/cache/store/disk.py
@@ -45,7 +45,7 @@ class OnDiskCacheStore(CacheStore):
 
     def get_cache_file_path(self, request: HttpParser, create: bool = False) -> str:
         request_method = text_(request.method)
-        request_host = text_(request.host)
+        request_host = text_(request.host if request.host else request.headers[b'host'][1])
         request_path = text_(request.path)
         request_body = sha512(request.body).hexdigest() if request.body else 'None'
 
--- a/proxy/plugin/cache/store/disk.py
+++ b/proxy/plugin/cache/store/disk.py
@@ -26,6 +26,8 @@ class OnDiskCacheStore(CacheStore):
     def __init__(self, uid: UUID, cache_dir: str) -> None:
         super().__init__(uid)
         self.cache_dir = cache_dir
+        if not os.path.isdir(self.cache_dir):
+            os.mkdir(self.cache_dir)
         self.cache_file_path: Optional[str] = None
         self.cache_file: Optional[BinaryIO] = None
 
--- a/proxy/plugin/cache/store/disk.py
+++ b/proxy/plugin/cache/store/disk.py
@@ -10,6 +10,7 @@
 """
 import logging
 import os
+import uuid
 import multiprocessing
 from typing import Optional, BinaryIO
 from hashlib import sha512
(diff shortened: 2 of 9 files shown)


abhinavsingh:
Thanks for the PR.  Working on a release this week at my end.  I'll review the PR later this week or over the weekend.  Cheers!!!

pasccom:
OK, I am working on a mode where the plugin does not contact upstream (even when intercepting TLS or if request is not cached). This will be very useful for automatic tests using proxy.py. I am mostly done. I only need to write tests for response when request is not cached. I will send another PR for this feature.

abhinavsingh:
@pasccom apologies for no review this week, unfortunately I am still working towards the release at my end.  It's unlikely I will be able to review it this week.  But will surely jump onto it over the weekend sometime.  Thank you!!!

proxy/plugin/cache/base.py
```diff
@@ -27,34 +32,109 @@ class BaseCacheResponsesPlugin(HttpProxyBasePlugin):
...
             self, request: HttpParser) -> Optional[HttpParser]:
+        self.__enabled = self.__class__.enabled.is_set()
+        if not self.__enabled:
+            return request
+
         assert self.store
+        logger.info("Upstream connexion %s:%d %s" %
+                    (text_(request.host), request.port if request.port else 0, text_(request.path)))
+
+        if request.port == 443:
```
abhinavsingh, line 70: We can perform TLS handshake with the client without connecting with upstream.  See `utils/wrap_socket` utility here https://github.com/abhinavsingh/proxy.py/blob/develop/proxy/common/utils.py#L154-L166 . We'll have to re-use server side certificates generated previously during TLS interception for handshake with the client.  Wdyt?

proxy/plugin/cache/base.py
```diff
@@ -27,34 +32,109 @@ class BaseCacheResponsesPlugin(HttpProxyBasePlugin):
...
+        assert self.__enabled is not None
+        if not self.__enabled:
+            return request
+
         assert self.store
-        return self.store.cache_request(request)
+        logger.info("Client request %s:%d %s" %
+                    (text_(request.host), request.port if request.port else 0, text_(request.path)))
+
+        if request.port == 443:
```
abhinavsingh, line 91: Assuming we have already completed TLS handshake with the client above in `before_upstream_connection` callback, here we can return response from the cache.  There are two possibilities:

1. Save unencrypted response as cache and encrypt it every time while serving from cache.  In-fact, encryption will be automatic when serving clear text response over the TLS connection.
2. Save encrypted response as cache.  Decrypt to parse it as necessary.  Finally serve data over established TLS connection.  This step involves an edge case that can occur when generated server side certificate expires, resulting into inability to decrypt the stored response or in worst case resulting into serving garbage data to client.

proxy/plugin/cache/store/disk.py
```diff
@@ -22,20 +25,70 @@
...
+    lock = multiprocessing.Lock()
 
     def __init__(self, uid: UUID, cache_dir: str) -> None:
         super().__init__(uid)
         self.cache_dir = cache_dir
-        self.cache_file_path: Optional[str] = None
+        if not os.path.isdir(self.cache_dir):
+            os.mkdir(self.cache_dir)
+        logger.debug("Opening cache list")
+        self.cache_list = open(os.path.join(self.cache_dir, 'list.txt'), 'at')
```
abhinavsingh, line 36: Going forward we must remove hardcoded `list.txt`.  This ideally should be user provided.  Please leave a `TODO` for now or address it by reading the path from an environment variable e.g. `PROXY_PY_CACHE_FILENAME`.  When unavailable, fallback to `list.txt`.

tests/plugin/test_http_proxy_plugins.py
```diff
@@ -56,6 +59,14 @@ def setUp(self,
             flags=self.flags)
         self.protocol_handler.initialize()
 
+    def tearDown(self) -> None:
+        tmpDir = Path(tempfile.gettempdir())
+        for f in tmpDir.glob('proxy-cache-*'):
+            if f.is_file():
+                os.remove(f)
+        if tmpDir.joinpath('list.txt').is_file():
```
abhinavsingh, line 67: During parallel testing, `os.remove` results in a deadlock as `list.txt` is also being used by other tests. See https://github.com/abhinavsingh/proxy.py/pull/425/checks?check_run_id=1197597223

abhinavsingh (changes requested):
Hi @pasccom .  Apologies for no show for about a month, I was swamped with releases at my end.  Getting back to this PR, I left a few comments.  Also, when I locally ran tests, `tests/testing/test_embed.py::TestProxyPyEmbedded::test_proxy_no_vcr` failed.

Overall, I think this PR is an excellent first step.  I would love to take this forward.  Let's not merge this PR into `develop` branch but instead into a branch named `cache-server`.  We can iterate over cache server implementation, make changes to the core as necessary before merging into `develop` branch.  Wdyt?

```
=================================================================================================== FAILURES ====================================================================================================
_____________________________________________________________________________________ TestProxyPyEmbedded.test_proxy_no_vcr _____________________________________________________________________________________

self = <tests.testing.test_embed.TestProxyPyEmbedded testMethod=test_proxy_no_vcr>

    def test_proxy_no_vcr(self) -> None:
        self.make_http_request_using_proxy()
    
>       self.assertIsNone(self.waitCached(Path(tempfile.gettempdir()) / 'list.txt',
                                          'GET', 'localhost', '/', 'None', 2))
E       AssertionError: 'proxy-cache-afd0b9587e804d25a50651812095c216' is not None

tests/testing/test_embed.py:117: AssertionError
--------------------------------------------------------------------------------------------- Captured stderr setup ---------------------------------------------------------------------------------------------
2020-10-02 13:33:26,846 - pid:25384 [I] load_plugins:541 - Loaded plugin proxy.http.proxy.HttpProxyPlugin
2020-10-02 13:33:26,846 - pid:25384 [I] load_plugins:541 - Loaded plugin proxy.http.server.HttpWebServerPlugin
2020-10-02 13:33:26,846 - pid:25384 [I] listen:71 - Listening on 0.0.0.0:61304
2020-10-02 13:33:26,849 - pid:25384 [D] start_workers:88 - Started acceptor#0 process 25418
2020-10-02 13:33:26,849 - pid:25384 [I] start_workers:94 - Started 1 workers
---------------------------------------------------------------------------------------------- Captured log setup -----------------------------------------------------------------------------------------------
INFO     proxy.common.flags:flags.py:541 Loaded plugin proxy.http.proxy.HttpProxyPlugin
INFO     proxy.common.flags:flags.py:541 Loaded plugin proxy.http.server.HttpWebServerPlugin
INFO     proxy.core.acceptor.pool:pool.py:71 Listening on 0.0.0.0:61304
DEBUG    proxy.core.acceptor.pool:pool.py:88 Started acceptor#0 process 25418
INFO     proxy.core.acceptor.pool:pool.py:94 Started 1 workers
============================================================================================ short test summary info ============================================================================================
FAILED tests/testing/test_embed.py::TestProxyPyEmbedded::test_proxy_no_vcr - AssertionError: 'proxy-cache-afd0b9587e804d25a50651812095c216' is not None
========================================================================================= 1 failed, 134 passed in 8.67s =========================================================================================
--- Logging error ---
  File "/usr/local/Cellar/python@3.8/3.8.3/Frameworks/Python.framework/Versions/3.8/lib/python3.8/logging/__init__.py", line 1084, in emit
    stream.write(msg + self.terminator)
  File "/Users/abhinavsingh/Dev/proxy.py/venv38/lib/python3.8/site-packages/_pytest/capture.py", line 424, in write
    return self.buffer.write(s.encode(self.encoding, "replace"))
ValueError: I/O operation on closed file
Call stack:
  File "/Users/abhinavsingh/Dev/proxy.py/proxy/plugin/cache/store/disk.py", line 41, in __del__
    logger.debug("Closing cache list")
Message: 'Closing cache list'
Arguments: ()
--- Logging error ---
  File "/usr/local/Cellar/python@3.8/3.8.3/Frameworks/Python.framework/Versions/3.8/lib/python3.8/logging/__init__.py", line 1084, in emit
    stream.write(msg + self.terminator)
  File "/Users/abhinavsingh/Dev/proxy.py/venv38/lib/python3.8/site-packages/_pytest/capture.py", line 424, in write
    return self.buffer.write(s.encode(self.encoding, "replace"))
ValueError: I/O operation on closed file
Call stack:
  File "/Users/abhinavsingh/Dev/proxy.py/proxy/plugin/cache/store/disk.py", line 41, in __del__
    logger.debug("Closing cache list")
Message: 'Closing cache list'
Arguments: ()
make: *** [lib-test] Error 1
```

abhinavsingh:
Will go ahead and merge this under `cache-server` branch.  Want to iterate over your work this weekend.  Thank you again!!!

proxy/plugin/cache/base.py
pasccom: Yes, I already looked in this direction. This is what I do locally for automatic testing purposes. I pushed my work on a branch on my fork https://github.com/pasccom/proxy.py/tree/develop_cache_no_upstream
I did not create a PR yet because I wanted this one first to be fixed before going further. The major issue with not contacting upstream is that I ensure `do_connect` is falsy in https://github.com/abhinavsingh/proxy.py/blob/8cc349be488bd8fc5c23dea4f1f83c8d0c02a6f0/proxy/http/proxy/server.py#L337-L338 and consequently `self.server` is `None` at https://github.com/abhinavsingh/proxy.py/blob/8cc349be488bd8fc5c23dea4f1f83c8d0c02a6f0/proxy/http/proxy/server.py#L279

proxy/plugin/cache/base.py
pasccom: I would prefer we use solution 1. It seems more practical to me. I sometimes had to look at the cache to understand what data got into it (upsteam has sporadic faults).
That is what I used on https://github.com/pasccom/proxy.py/tree/develop_cache_no_upstream

proxy/plugin/cache/store/disk.py
pasccom: Yes, I agree, I will add a `TODO`.

tests/plugin/test_http_proxy_plugins.py
pasccom: Ah OK! I did not know CI was using parallel testing. When I run them locally, I think they execute sequentially. Am I right?
I tried to investigate this issue but I could not understand what was happening (I am under Linux and have no ways of running Windows test locally). 

Actually, multi-threading is already a major issue with the disk-based cache, since I have to find a way to synchronize file accesses. I used a `multiprocessing.Lock()`, but I am not sure it is the best solution out there. Indeed, the reader can access the file simultaneously, while the writes must access the file sequentially. A "read-write lock" would be more appropriate. Or we could go for OS file locks, but that would be platform-dependent.
For multi-processing, I am not sure what we can do, since `multiprocessing.Lock()` will be insufficient, as it is not aware of the other instances of tests running in parallel, right?

pasccom:
Hello @abhinavsingh! Nice to have news, I feared you got a bad COVID.
Yes, there is still some work to do, especially regarding multi-threading and multi-processing. 
I think there is an issue with `Transfer-Encoding: chunked` as well, I got it a few days ago on my prototype. I did not had time to check if it works or not with this version. I just removed the header which was set by the server on my prototype.
I will let you work over the weekend and come to see what you did. Hopefully, I will have some time to fix the remaining issues next week.
Best regards,

proxy/plugin/cache/base.py
abhinavsingh: Correct, `before_upstream_connection` callback can return `None` to avoid connecting to upstream.  Internally it will keep `do_connect = False`.

And thanks for bringing this to notice, `on_client_data` does ensures server connection before proceeding.  Might have to revisit the request state diagram to clean up some of these bits/expectations.

proxy/plugin/cache/base.py
abhinavsingh: +1 agreed.  Privacy can be argued with solution 1 but this aspect can be controlled by using proper ACL/Permissions over stored cache directories or files.

proxy/plugin/cache/store/disk.py
abhinavsingh: jFYI, yesterday I merged https://github.com/abhinavsingh/proxy.py/pull/438 which will soon allow plugins to define custom flags.  Will be migrating configurable parameters within various plugins _(e.g. --blacklist-ip-addresses, --cache-dir, ....)_ as flag this week.  We should be able to re-use this mechanism for `list.txt`.

Talking of `list.txt`, at some point in future far away :) we must also revisit cache schema to ensure best performance under various scenarios _(long running instances, gigabytes of cache data, ....)_

tests/plugin/test_http_proxy_plugins.py
abhinavsingh: > Ah OK! I did not know CI was using parallel testing. When I run them locally, I think they execute sequentially. Am I right?

Yes, tests run sequentially.  Thinking about it further, as same command invocation is used for CI, there is no reason they should run in parallel on GitHub actions.

> For multi-processing, I am not sure what we can do, since multiprocessing.Lock() will be insufficient, as it is not aware of the other instances of tests running in parallel, right?

Yes.  And idea of shared information about other tests doesn't sound right in the first place.  There are some sporadic issues related to `multiprocessing` and `windows` too.  I have a windows vm locally, will give it a try.

We can bypass tests conditionally. Example

```
@unittest.skipIf(
    os.name == 'nt',
    'Open file limit tests disabled for Windows')
class TestSetOpenFileLimit(unittest.TestCase):
```

> Actually, multi-threading is already a major issue with the disk-based cache, since I have to find a way to synchronize file accesses.

Yes.  A single file will require synchronization.  Avoiding lock/synchronizations can also boost the cache storage performance by many folds.  `list.txt` is essentially an index file.  We can even rename it appropriately in later iterations.  It contains metadata about stored item in a random order i.e. stored indexes are not ordered by any attribute.  We can also see `list.txt` as serialized index store as indexes exists in the order of requests.

Currently, we don't have an update flow i.e. once written, cache is immutable and never updated.  But, design wise we must prepare for an update cycle :) where cache is updated either based upon time based expiry, query parameters etc.  This decision making can be moulded into cache plugin lifecycle callback.

> I used a `multiprocessing.Lock()`, but I am not sure it is the best solution out there. Indeed, the reader can access the file simultaneously, while the writes must access the file sequentially. A "read-write lock" would be more appropriate. Or we could go for OS file locks, but that would be platform-dependent.

The task of on disk based storage leads us into database designing wonderland :) Good thing is our cache storage is pluggable and we can easily add new storage mechanisms _(leveldb, sqlite, cloud storages, ....)_

Few thoughts around approaching disk based storage using zero-dependency philosophy:

1) Storage:  We are currently caching responses as a single file per request.  So storage becomes pretty straightforward.  But for a long running server or with lots of data, I do expect us to run into situations where OS starts to run out of file descriptors _(possibly fixable by tweaking kernel parameters)_.  Let's not bother about it for now as fixing this is relatively easier if we get our indexes right.

2) Indexes:  Without any ordering on stored indexes, we indirectly increase one or both:
a) Startup time:  Cache plugin is busy ordering the indexes in-memory (if we decide to pre-warm the cache index).
b) Lookup time:  Cache plugin is busy iterating over unordered indexes (if we decide to consult index file for every read).

In short, ordering (fully or partially) indexes on disk can greatly benefit overall performance.  Thinking of primitive ways :) maintaining one index file per alphabet e.g. `a-list.txt`, `b-list.txt`, .... where each file store data about domains starting with respective alphabet (or any unicode character).  Such partitioning strategies can greatly reduce the lock contention across workers.

At some point we can also use an in-memory cache e.g. LRU cache to minimize disk hits.  Ideally, I would like to have a running copy of indexes in-memory but that is not always practical with lots of indexes (and not enough available RAM).

3) Synchronization: There is in-build `pubsub` mechanism in `proxy.py`.  Try it independently here https://github.com/abhinavsingh/proxy.py/blob/develop/examples/pubsub_eventing.py.

For some background, eventing mechanism was developed for `devtools` integration.  This allows `proxy.py` traffic to be inspected using `chrome devtools`.  `devtools` expects a `websocket` connection sending stream of events as described in `devtools` protocol specification.  In short, we do have a mechanism to produce a single stream of custom events from various worker processes.  Most importantly, eventing mechanism relieves us from synchronization pains.  

An idea could be to simply publish cache payload as events from within cache plugin.  Then, a single global subscriber thread or multiple subscriber threads (one per worker process) can consume events and write indexes to disk either synchronously (as events are received) or in batches.

Hopefully, we can mix some of the above ideas to build our own storage layer :)

abhinavsingh:
> Hello @abhinavsingh! Nice to have news, I feared you got a bad COVID.

Thank you :) I am doing fine, simply got super busy with other releases.

> Yes, there is still some work to do, especially regarding multi-threading and multi-processing.
> I think there is an issue with `Transfer-Encoding: chunked` as well, I got it a few days ago on my prototype. I did not had time to check if it works or not with this version. I just removed the header which was set by the server on my prototype.

Totally chunked encoding is another problem we must resolve.  This is probably an indication that cache must be written only after the connection has been closed or when request is complete.

proxy/plugin/cache/base.py
```diff
@@ -27,34 +32,109 @@ class BaseCacheResponsesPlugin(HttpProxyBasePlugin):
...
+            if enabled:
+                self.enabled.set()
+
+        def __get__(self, obj: Optional[object], owner: type) -> Any:
+            if obj is None:
+                return self.enabled
+            return None
+
+    # Dynamically enable / disable cache
+    enabled = EnabledDescriptor(True)
```
abhinavsingh, line 47: Also must note that dynamic `enabled` flag doesn't work on `MacOS` `Python 3.8`.  https://github.com/abhinavsingh/proxy.py/pull/425/checks?check_run_id=1197597150

I am also able to reproduce it locally.  Somehow the event changes are not propagated within plugin method checks.

tests/plugin/test_http_proxy_plugins.py
pasccom: > We can bypass tests conditionally. Example
> 
> ```
> @unittest.skipIf(
>     os.name == 'nt',
>     'Open file limit tests disabled for Windows')
> class TestSetOpenFileLimit(unittest.TestCase):
> ```

Yes, I knew we can do this, but I wanted you to see if you better understands what happens on Windows. The current output from the tests is a bit short and I did not want to push tests commits only to see a more verbose output.

> Yes. A single file will require synchronization. Avoiding lock/synchronizations can also boost the cache storage performance by many folds. `list.txt` is essentially an index file. We can even rename it appropriately in later iterations. It contains metadata about stored item in a random order i.e. stored indexes are not ordered by any attribute. We can also see `list.txt` as serialized index store as indexes exists in the order of requests.

Exactly, that is how I designed it.
 
> Currently, we don't have an update flow i.e. once written, cache is immutable and never updated. But, design wise we must prepare for an update cycle :) where cache is updated either based upon time based expiry, query parameters etc. This decision making can be moulded into cache plugin lifecycle callback.

Yes, I kept that for later, because I think that there is quite a large amount of work to implement all cache expiracy features.

> The task of on disk based storage leads us into database designing wonderland :) Good thing is our cache storage is pluggable and we can easily add new storage mechanisms _(leveldb, sqlite, cloud storages, ....)_
> 
> Few thoughts around approaching disk based storage using zero-dependency philosophy:
> 
> 1. Storage:  We are currently caching responses as a single file per request.  So storage becomes pretty straightforward.  But for a long running server or with lots of data, I do expect us to run into situations where OS starts to run out of file descriptors _(possibly fixable by tweaking kernel parameters)_.  Let's not bother about it for now as fixing this is relatively easier if we get our indexes right.

I do not think we will run out of file descriptors because there is only one open file per non-completed request plus the index. This should not be too many for the OS, unless we open a lot of requests in parallel on purpose.
Though, if we use it for a long time we will run into the limit of inodes per filesystem. For Ext2/3/4 filesystems, it can be configured when formatting the disk. To avoid this we can use subfolders as is done in e.g. `.git/objects` "cache". 

> 2. Indexes:  Without any ordering on stored indexes, we indirectly increase one or both:
>        a) Startup time:  Cache plugin is busy ordering the indexes in-memory (if we decide to pre-warm the cache index).
>        b) Lookup time:  Cache plugin is busy iterating over unordered indexes (if we decide to consult index file for every read).
> 
> 
> In short, ordering (fully or partially) indexes on disk can greatly benefit overall performance. Thinking of primitive ways :) maintaining one index file per alphabet e.g. `a-list.txt`, `b-list.txt`, .... where each file store data about domains starting with respective alphabet (or any unicode character). Such partitioning strategies can greatly reduce the lock contention across workers.

Yes, this will become an issue for large caches. The list by alphabet letter seems a good compromise, but we have to get a reliable way to determine it, without having everything in `w` (many websites still use `www`). Or we could even go for a per-host index file. I do not know how fast is `os.path.is_file()`.

PR #534, opened 2021-04-07
Add header plugin

sam-kleiner:


sam-kleiner:
Still need to write some tests, will look into that later if you want them

proxy/plugin/add_header.py
```diff
@@ -0,0 +1,57 @@
...
+    :license: BSD, see LICENSE for more details.
+"""
+import re
+from typing import Optional
+
+from ..common.flag import flags
+from ..http.parser import HttpParser
+from ..http.proxy import HttpProxyBasePlugin
+
+DEFAULT_HEADER = "X-Proxy-Py:Proxied with proxy.py"
```
abhinavsingh, line 18: I think we don't need any default headers from a plugin.  This will be little surprising for end user who has no idea about default headers added by the plugin.  Wdyt?

proxy/plugin/add_header.py
```diff
@@ -0,0 +1,57 @@
...
+    def before_upstream_connection(
+            self, request: HttpParser) -> Optional[HttpParser]:
+        return request
+
+    def handle_client_request(
+            self, request: HttpParser) -> Optional[HttpParser]:
+        if not self.flags.add_header:
+            self.flags.add_header = [DEFAULT_HEADER]
+
+        for header in self.flags.add_header:
```
abhinavsingh, line 46: Yes please add a test for this block, specially we want to check for following cases:

1. Single `--add-header` works fine
2. Multiple `--add-header` works fine
3. `--add-header` with `:` and other special characters work as expected

abhinavsingh (commented):
Thank you @sam-kleiner , I like this plugin, a simple way to add custom headers.  Can you please add tests also, I mentioned a few key cases in the comments.

abhinavsingh:
@sam-kleiner Closing the PR for now.  Let's re-open once we have addressed the review comments.  Thank you!!!

PR #553, opened 2021-05-16
`--plugin` and `--plugins` flags doesn't work consistently together

sakurai-youhei:
**Issue** Custom flags in my own plugin won't be discovered with `--plugins`. Workaround is using `--plugin` instead.

- Doesn't work: `proxy --plugins my.own.Plugin --custom-flag spam`
- Does work: `proxy --plugin my.own.Plugin --custom-flag spam`

**Singular v.s. Plural** `--plugin` sounds consistent but `--plugins` is being used intentionally.

**Proposal** Fix the issue without removing the original singular `--plugin`.


abhinavsingh:
@sakurai-youhei Thank you for the PR.  Indeed plurals should work too.  However, tests seems to be failing.  Example: https://github.com/abhinavsingh/proxy.py/pull/553/checks?check_run_id=2595396247. Can you please take a look at them.  You can run them locally and debug too.  Thank you!!!

sakurai-youhei:
@abhinavsingh Thank you for pointing it out; It's my fault as I didn't imagine such small change breaks anything.

Now that I makes the change with writing small tests. Hope they would be compliant with existing manners and pass the checks. :)

abhinavsingh:
@sakurai-youhei My apologies, but I highly recommend you setup pre-commit hooks locally , see https://github.com/abhinavsingh/proxy.py#setup-pre-commit-hook. You can also manually just run `make` before pushing to remote, this will run all git workflows on your local system.

This will avoid any unnecessary failures which you can catch locally first.  Looks like we have a lint error next.

Screen Shot 2021-05-17 at 7 14 30 PM

sakurai-youhei:
@abhinavsingh I am sorry for my laziness again. I set it up and confirmed no errors locally.

sakurai-youhei:
@abhinavsingh Let me re-work it as I didn't check the change with older version of Python...

abhinavsingh:
@sakurai-youhei This should be fixed after

- #725

sakurai-youhei:
Thanks!

PR #554, opened 2021-05-16
64 run uniswapv2 core tests from proxy ci

vasiliy-zaznobin:
Add a uniswap-v2-core test to check


vasiliy-zaznobin:
wrong destination

PR #605, opened 2021-09-11
Fixed argument passing & hide shells when executing openssl commands

hash3liZer:
There are 2 fixes in this pull request: 

) The logs related arguments weren't passable from command line version. The pull requests fixes that issue. 
) If you use this tool in a gui related tool, when decrypting traffic, you will see command prompt or related shell environment being popped up again and again. This is due to creation flags when calling subprocess.Popen. This pull request hides the shell when executing command prompts for GUI Mode.


proxy/proxy.py
```diff
@@ -184,6 +184,9 @@ def initialize(input_args: Optional[List[str]]
             sys.exit(0)
 
         # Setup logging module
+        args.log_file  = cast(Optional[str], opts.get('log_file', args.log_file))
+        args.log_level  = cast(Optional[str], opts.get('log_level', args.log_level))
+        args.log_format  = cast(Optional[str], opts.get('log_format', args.log_format))
```
abhinavsingh, line 189: These flags already exists.  See https://github.com/abhinavsingh/proxy.py#flags

Can you expand why exactly are you trying to add them here again.

Thank you!!!

PS: Apologies for delay, been very tied up with other projects.

hash3liZer: Yes, sure. Ok, these flags work perfect when supplied from command line. For example, this works perfect: 

```
proxy --log-file /tmp/filer.txt ....
```

However, if you do the same in Python:
```python
import proxy
proxy.main(
    hostname=...,
    port=8080,
    log_file="/tmp/filer.txt"
)
```

It doesn't work. The above statement fixes that issue.

hash3liZer: However, i just realized now that linux & mac doesn't startupinfo parameter for subprocess module. I guess, we will have to modify the pull request.

abhinavsingh:
@hash3liZer Unsure if this is review/merge ready.  IIUC, we needed to revisit this PR.  Closing it for now.  Feel free to re-open if my understanding is not correct.  Thank you

PR #635, opened 2021-10-26
Fix client request dropped

sam01101:
After some playing with `proxy.py`, I've found out that most of the time `SSLWantReadError` is raised and the request is dropped,

I've looked into the document and found out that this happened is because it don't have enough data to receive.

This patch fixes this problem by doing a while loop on `select.select` until success.

Ref (And the original code): https://docs.python.org/3.8/library/ssl.html#ssl-nonblocking

Q: There's already a retry mechanism by doing `return False`, why you need to add a while loop?
A: See above.

Q: Won't while loop be stuck / slows down the performance?
A: It's hard to say, at least it doesn't seem like blocking for me, most of the problem is `the data is not enough`.

Quote from official doc:
> Calling `select()` tells you that the OS-level socket can be read from (or written to), 
> but it does not imply that there is sufficient data at the upper SSL layer.
> For example, only part of an SSL frame might have arrived. Therefore,
> you must be ready to handle `SSLSocket.recv()` and `SSLSocket.send()` failures, 
> and retry after another call to `select()`.

Update: Should we just add a retry counter and let the user to decide times to retry? Will that solve the while loop problem? Is there a better way to check whether the socket is ready to read or not?

Extra info:

<details>
<summary>Retry can be spammy</summary>

Code:

image

Log:

image

</details>


abhinavsingh:
@sam01101 I think looping might not be the right fix here.  In worst case it can block the entire event loop.  I am thinking, as per the description here https://docs.python.org/3/library/ssl.html#ssl.SSLWantReadError

> exception ssl.SSLWantReadError
>> A subclass of SSLError raised by a non-blocking SSL socket when trying to read or write data, but more data needs to be received on the underlying TCP transport before the request can be fulfilled.

We should not drop the request, instead just wait for the next read event to be available, where likely more data will be available.  Wdyt?

TL;DR -- We must retry, but by against registering for read events and waiting for next read callback.  While loop may block in worst case, as more data might not be immediately available.

abhinavsingh:
> How to duel with flake8 `C091`?

You can either refactor the method to reduce complication or just bump the max complexity value to 21

abhinavsingh:
@sam01101 I looked closely into the code and retries must automatically happen.  Notice how `handle_readables` returns `False` for `SSLWantReadError` scenario.  Returning `False` simply means, retry again for reads.  If we look at `get_events`, we can see `self.client.connection` is always registered for read events.

Hence, I believe, if the connection gets dropped, the event is likely triggered from upstream server side.

TL;DR -- Your original question about, "should we add retries?".  It's already baked in.

abhinavsingh:
One way would be, simply verify this scenario using tests.   Mock upstream endpoint can simulate `SSLWantReadError` within the code and later send more data or just drop the connection.  At-least we can then verify how `proxy.py` behaves under such circumstances.

sam01101:
Thanks for the analyze, but sorry I'm busy irl so I'll test it with test case when I'm free

abhinavsingh:
> Thanks for the analyze, but sorry I'm busy irl so I'll test it with test case when I'm free

No worries.  Let's revisit this when we can.  Currently I think this is not an issue and connections are only closed when client or upstream server itself drops the connection.

I have raised an issue to add tests for missing clauses within `server.py` which will also cover `SSLWantRead` scenarios.  At least we can be rest assured that library is working as expected.

PR #641, opened 2021-10-30
Fix the `typing_extensions` runtime dependency version

webknjaz:
This patch updates the code to only use `typing-extensions` when the stdlib type is unavailable. It also makes it unpinned in the package metadata (keeping the pins in the test env requirements). This allows it to be used in virtualenvs along with other requirements that have dependencies on `typing-extensions` too.

### DIFF
--- a/proxy/common/types.py
+++ b/proxy/common/types.py
@@ -10,10 +10,17 @@
 """
 import queue
 import ipaddress
+import sys
 
 from typing import TYPE_CHECKING, Dict, Any, List, Union
 
-from typing_extensions import Protocol
+# NOTE: Using try/except causes linting problems which is why it's necessary
+# NOTE: to use this mypy/pylint idiom for py36-py38 compatibility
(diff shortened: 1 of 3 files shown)


proxy/common/types.py
```diff
@@ -10,10 +10,17 @@
...
 from typing import TYPE_CHECKING, Dict, Any, List, Union
 
-from typing_extensions import Protocol
+# NOTE: Using try/except causes linting problems which is why it's necessary
+# NOTE: to use this mypy/pylint idiom for py36-py38 compatibility
+# Ref: https://github.com/python/typeshed/issues/3500#issuecomment-560958608
+if sys.version_info >= (3, 8):
+    from typing import Protocol
+else:
+    from typing_extensions import Protocol
```
abhinavsingh, line 23: Thank you :). Was totally unnecessary to include typings for all future Python versions

PR #652, opened 2021-11-02
Add a no-op check job to GHA for branch protection

webknjaz:
This patch adds an "empty" job that depends on all the important ones
making it possible to use just this `check` job in the branch
protection settings. This reduces the maintenance burden by preventing
the need to update these settings on any changes to the job
declarations.


webknjaz:
@abhinavsingh with this, you can replace individual jobs from the `Library - *` workflow with just `check` in the branch protection settings.

PR #654, opened 2021-11-02
Add a config for flake8

webknjaz:
Why have a config separate and explict? This makes maintaining its settings easier. It's impossible to keep all the settings as CLI args, especially in multiple places. It makes it easier to share the same configs when the tool is invoked from different places or just called directly.
Why not dump it in one of the other configs? It's hard to locate settings from different tools in one file, especially when it's big. Plus it's more obvious when you try to find which file the config is in after a while when you forget where it is. In some projects you'd have to open like 4-5 different files trying to locate the config...


abhinavsingh:
Really appreciate these code quality related PR +10000

abhinavsingh:
@webknjaz Indeed.  Thank you Sviatoslav.  Frankly, I myself need to catch up with some of the changes done.  Your PRs have opened up a whole new direction into what's possible with these rules and how to maintain top-notch code quality.

I am definitely going to spend more time digging into them.  And totally makes sense to ramp up new contributors via related issues.

PR #658, opened 2021-11-02
Make the YAML files follow consistent style

webknjaz:
This patch enables linting of the previously inconsistent files and adjusts them to conform the same style.

This PR is built on top of https://github.com/abhinavsingh/proxy.py/pull/657. Merge it and its deps first.

PR #659, opened 2021-11-03
chore: Simplify `if` expression by using `or`

yezz123:


proxy/common/types.py
```diff
@@ -8,6 +8,7 @@
     :copyright: (c) 2013-present by Abhinav Singh and contributors.
     :license: BSD, see LICENSE for more details.
 """
+
```
webknjaz, line 11: This is clearly unrelated to changing the if-clause. It's strongly recommended not to mix style changes with the functional ones. Besides, it's best to ask the maintainer if they even want those in the first place. Also, looking at other modules, they actually consistently don't have this extra empty line here.
```suggestion
```

webknjaz (changes requested):
Hey, @yezz123! I've got a few comments on this change.

This PR title doesn't correspond to its contents. Could you update it to represent the effect of applying this patch to the `develop` branch? Also, it'd be useful to have the PR description filled out explaining the motivation behind this change.

Since there seem to be a number of refactoring pieces, it's probably best to add inline comments to each of these places in the `Files` tab explaining why you think this is needed.

I've also submitted a number of PRs setting up linting improvements in the CI, may I ask you to wait until @abhinavsingh merges them and then rebase this on top so that those checkers are taken into account?

P.S. FYI here's a nice article with pointers on how to compose great pull requests: https://mtlynch.io/code-review-love/.

proxy/common/types.py
abhinavsingh: @yezz123 Once #658 is in, we should be able to get this in.  Regarding extra line, I think it's not needed.

You'll also need to update your PR as currently `This branch is out-of-date with the base branch`.

Thank you

abhinavsingh:
@yezz123 This PR has again run into "This branch is out-of-date with the base branch".  Unsure why am I unable to "Update Branch" from within PR.  Could be because of permission issues on upstream fork?  Can you take a look.  Thank you

Screen Shot 2021-11-04 at 5 01 44 PM

abhinavsingh (changes requested):
PTAL

yezz123:
@abhinavsingh could you check now?

proxy/plugin/filter_by_url_regex.py
```diff
@@ -92,10 +89,6 @@ def handle_client_request(
                     headers={b'Connection': b'close'},
                     reason=b'Blocked',
                 )
-                # stop looping through filter list
-                break
```
webknjaz, line 96: This removes unreachable code. Could you also try dropping the linter exclusions at https://github.com/abhinavsingh/proxy.py/blob/7c5877b/.pylintrc#L129 and https://github.com/abhinavsingh/proxy.py/blob/7c5877b/.flake8#L149 to see if this was the only place with this type of a linting violation?

abhinavsingh:
> @abhinavsingh could you check now?

Can you please update again so that we can get your PR in.   Leave a TODO for linter checks as proposed by @webknjaz and we can revisit it outside of this PR.  Wdyt folks?

webknjaz:
> > @abhinavsingh could you check now?
> 
> Can you please update again so that we can get your PR in. Leave a TODO for linter checks as proposed by @webknjaz and we can revisit it outside of this PR. Wdyt folks?

I guess adding two lines with todo from the linter configs does not differ much from removing two lines. If the violations are fixed, why keep the ignores?

abhinavsingh:
> I guess adding two lines with todo from the linter configs does not differ much from removing two lines. If the violations are fixed, why keep the ignores?

Agreed, if PR author has know-how about underlying workflows *(not GHA but lint related)*, best to give it a try within the PR.   Having said that, I don't expect contributors to necessarily know where to tweak lint warnings or improve them.   A regular contributor probably should, but not necessary for first-time contributors.

TL;DR -- As a first time contributor if CI passes, task is done within that PR context.  Any other enhancements can be done in follow-up PRs.  Wdyt?

I am also considering writing a contributors guide.  We can state various pre-requisites for first-time and regular contributors.  Currently, I see first time contributors struggling with a green CI *(multiple back-and-forth)*.

webknjaz:
> TL;DR -- As a first time contributor if CI passes, task is done within that PR context. Any other enhancements can be done in follow-up PRs. Wdyt?

I guess that's fine. OTOH if the contributor is able to remove a specific line in a specific file, why not ask about it? I've linked the exact lines @ https://github.com/abhinavsingh/proxy.py/pull/659#discussion_r742804560. It's not that hard. If the contributor doesn't want to do this, they should state this explicitly, and this would be fine as well, I think.

abhinavsingh:
> they should state this explicitly, and this would be fine as well, I think.

I agree about the comm part

abhinavsingh:
Several of these recommendations made into the code as part of refactoring.  I have picked up on remaining changes including lint warning removal in #904

PR #660, opened 2021-11-03
v2.4.0

abhinavsingh:
# Bug Fixes
- Fix logging issues which impacted `>= Python 3.8` environments
- Fix an issue in `HttpParser` where `Content-Length: 0` header accompanied by `body` could result in an infinite loop
- `ProxyPool` updates to work reliably for `https` requests
   - Also, now `ProxyPool` is totally async.  This is one of the top request by several folks.  Thanks for for your patience and being persistent :)

# Enhancements

- Allow plugins to register for custom descriptors.
   - This allows plugins to perform async IO operations while sharing the loop with `proxy.py` core.
- Introduced `EventManager` which eases eventing core setup and teardown
- Now plugins can customize access log format using `on_access_log` plugin callback.
   - See `DEFAULT_HTTP_ACCESS_LOG_FORMAT` and `DEFAULT_HTTPS_ACCESS_LOG_FORMAT` for default values
- Decouple `AcceptorPool` and eventing core setup/teardown
- Expose actual server port via `proxy.pool.flags.port`.
   - This is required by folks who use `--port=0` to bind on an ephemeral port
- Several enhancements to code quality, styling, lint, CI workflows.  Thanks to @webknjaz


abhinavsingh:
Closing for now.  While open it unnecessarily keeps running workflows.

PR #661, opened 2021-11-03
Enable the `add-trailing-comma` pre-commit fixer

webknjaz:


abhinavsingh:
@webknjaz Enabled "pre commit" app, removed requirement on "pyup" for now.  Looks like `types-paramiko` was never made available to `Python 3+`, resulting in pre commit hook failures.  It's harmless, merging this PR.

In future, I already plan on removing `paramiko` in favor of `asyncssh`, so this will automatically go away.  In the meantime, will try to address this more cleanly.  Possibly by using a guard in requirements and no type check pragma for offending lines probably.

PR #662, opened 2021-11-03
Migrate the pytest invocation to `tox`

webknjaz:


webknjaz:
@abhinavsingh now is a good time to switch the branch protection to use the `check` job only for this workflow.

abhinavsingh:
A bite from our own snake :)

Screen Shot 2021-11-04 at 5 03 33 PM

webknjaz:
Ah, never mind. I've messed up that rebase but now it's fine :)

PR #666, opened 2021-11-04
Pre-install mypy deps in `pre-commit.ci`

webknjaz:
`pre-commit.ci` web-service runs checks in a container with networking disabled. This is why `mypy --install-types` doesn't work there and the additional dependencies need to be declared explicitly. This patch does just that.


webknjaz:
https://github.com/abhinavsingh/proxy.py/pull/667 fixes the other failure that appears in `pre-commit.ci`.

PR #667, opened 2021-11-04
Fix the YTT201 flake8 violation

webknjaz:
This approach makes sure that if there's ever Python 4, it wouldn't
fall back to the Python 2 behavior silently.


webknjaz:
@abhinavsingh with this merged, you can now add `pre-commit.ci` to required checks in the branch protection, if you wish.

PR #680, opened 2021-11-05
Add a config for pytest

webknjaz:


abhinavsingh:
@webknjaz Anything that this PR could do to break `e2e` for `3.10` on `ubuntu`?

https://github.com/abhinavsingh/proxy.py/runs/4120221951?check_suite_focus=true#step:9:230

Unsure why are we getting these errors all of a sudden.  Are you encountering them locally?  On my macOS things seem ok.  Looks Ubuntu specific.

This check failed but I somehow missed it and merged the PR.

webknjaz:
I don't think so. Maybe GH updated their Pythons or VM images today in between the merges?

webknjaz:
Or maybe some of the transitive dependencies got updated?

webknjaz:
Either way, I think it's easy to check this theory by going to the previous commit's workflow runs and restarting the job. If that fails, we could be sure that it's unrelated.

webknjaz:
That job you linked is not e2e. e2e got canceled with the rest of the workflow because of the newer workflow run on the same PR so it's not indicative.

abhinavsingh:
> That job you linked is not e2e. e2e got canceled with the rest of the workflow because of the newer workflow run on the same PR so it's not indicative.

Yes, `e2e` was cancelled.  Its `lib` which is failing on Ubuntu.

Indeed could be due to minor version or image update at GHA end.  Let me re-run previous workflows again to confirm the theory.

webknjaz:
@abhinavsingh I've figured it out. You were right, it's indeed caused by this PR. A fix incoming.
Long story short, I turned on `filterwarnings` to make any warnings into errors, per best practices and it's caught a new deprecation warning.

abhinavsingh:
@webknjaz I tried re-running https://github.com/abhinavsingh/proxy.py/actions/runs/1424042850 .  It's a day older and tests worked out fine.  Started https://github.com/abhinavsingh/proxy.py/actions/runs/1427499665 , if this fails, might have to dig deeper.

CI history says this PR started breaking `3.10` on `ubuntu`, which is weird because this PR only introduces a config change.

webknjaz:
Here you go: https://github.com/abhinavsingh/proxy.py/pull/683.

PR #683, opened 2021-11-05
Ignore a `DeprecationWarning` in `pytest` config

webknjaz:
This was added in Python 3.10 and needs to be dealt with properly.

Action items:
* [ ] Stop calling `asyncio.get_event_loop()` outside of a running event loop
* [ ] Replace it with `asyncio.get_event_loop_policy().get_event_loop()` maybe

### DIFF
--- a/pytest.ini
+++ b/pytest.ini
@@ -56,6 +56,9 @@ faulthandler_timeout = 30
 filterwarnings =
   error
 
+  # FIXME: Address the deprecation warning in Python 3.10 and revert this ignore
+  ignore:There is no current event loop:DeprecationWarning:proxy.core.acceptor.threadless
+
 junit_duration_report = call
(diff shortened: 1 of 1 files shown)


abhinavsingh:
Thank you, will send out proper fix as highlighted.

abhinavsingh:
Looked into it, per comment `get_event_loop_policy().get_event_loop()` will be returned if none exists.  Irrespective, I have updated the code in threadless module to directly use `get_event_loop_policy`, doesn't seem to break anything apart :)

Screen Shot 2021-11-06 at 3 39 49 PM

Ideally, as a consumer of event loop API, we must get an event loop when requested, either from current context or a new one automatically.

IIUC, this is an issue only on `Ubuntu` currently.

PR #685, opened 2021-11-06
Fix python 3.10 @ ubuntu pytest ignore

abhinavsingh:


pytest.ini
```diff
@@ -57,7 +57,7 @@ filterwarnings =
   error
 
   # FIXME: Address the deprecation warning in Python 3.10 and revert this ignore
-  ignore:There is no current event loop:DeprecationWarning:proxy.core.acceptor.threadless
+  # ignore:There is no current event loop:DeprecationWarning:proxy.core.acceptor.threadless
```
webknjaz, line 60: Why didn't you just delete these lines?

tests/test_circular_imports.py
```diff
@@ -1,22 +1,32 @@
-"""Tests for circular imports in all local packages and modules.
+# -*- coding: utf-8 -*-
```
webknjaz, line 1: This is unnecessary in py3-only codebase

pytest.ini
abhinavsingh: Totally, I initially didn't delete coz I wasn't confident if this would work.  Later i missed removing them altogether before merge.

tests/test_circular_imports.py
abhinavsingh: Will remove within `README` PR now :)

webknjaz: You could have a `local` hook in the `pre-commit` config.

PR #692, opened 2021-11-06
Expose pre-commit tool hooks execution times

webknjaz:


abhinavsingh:
Thank you for surfacing timing info, it helps highlight the issue at hand.

Below is a screenshot for all other tasks `(except flake8, mypy, PyLint)`

Screen Shot 2021-11-07 at 4 27 26 AM

Below is the timing info for culprit tasks.  Collectively they are exhausting quotas faster than all other things combined across all workflows.  Not only are they running for `~51 seconds` out of `1 minute, 37.39 seconds` a.k.a `~50%` duration, they run at almost ~100% cpu usage across all available cores.

Note that, total time also includes environment setup steps.

Screen Shot 2021-11-07 at 4 28 29 AM

I also observe, `pre-commit.ci` didn't even execute the `PyLint` step for this PR.  Reference https://results.pre-commit.ci/run/github/12228178/1636238158.EAx8iiubSdWC9J0QOMYjTw

webknjaz:
> Below is the timing info for culprit tasks.  Collectively they are exhausting quotas faster than all other things combined across all workflows.  Not only are they running for `~51 seconds` out of `1 minute, 37.39 seconds` a.k.a `~50%` duration, they run at almost ~100% cpu usage across all available cores.
> 
> Note that, total time also includes environment setup steps.

As I mentioned in the other issue, GHA doesn't count the CPU load, just time the jobs were running the workload. Taking into account the number of the other test jobs and workflows + the fact that the minutes for macos workers are multiplied by 2 and the time consumed by the windows workers is multiplied by 10, I find it hard to believe that this linting job is a bottleneck.
I could imagine, though, that when a large number of jobs is requested, GH would throttle them. To fight this, I've added cancelation for the old jobs on PRs so that only one last workflow runs on the PR. But I only did that for "lib". It would help with the CI backlog to scale that patch to the other workflows (although I was going to try just moving all the jobs into a single workflow which would have the same effect).



> I also observe, `pre-commit.ci` didn't even execute the `PyLint` step for this PR.  Reference https://results.pre-commit.ci/run/github/12228178/1636238158.EAx8iiubSdWC9J0QOMYjTw .  May be because it is smart about identifying exhausting steps?

No, it's because I configured it this way. That hook relies on the externally installed pylint, this is solved by tox. But pre-commit.ci only runs what's in the pre-commit config, it cannot configure the env as necessary in other way.

PR #693, opened 2021-11-06
Work

abhinavsingh:


proxy/core/acceptor/acceptor.py
```diff
@@ -67,18 +73,26 @@ def __init__(
             event_queue: Optional[EventQueue] = None,
     ) -> None:
         super().__init__()
+        self.flags = flags
+        # Lock shared by all acceptor processes
```
webknjaz, line 77: Have you considered using attribute docstrings or `#: xxx` comments so that they could eventually get into the docs if/when Sphinx is added?

abhinavsingh: Yes, I do want to get to it.  Ideally, at some point I would love to see entire `README` or `doc` folder generated via docstrings.

Frankly, I just haven't settled down on what docstring standard to follow :)

PR #714, opened 2021-11-08
Add `--num-acceptors` flag + Allow `work_klass` via `Proxy` context manager kwargs

### LINKED ISSUE #706: [Acceptors] Decouple Threadless pool setup/teardown from acceptor pool
Currently, `--num-workers` is shared by both `acceptors` and `threadless` process pool.  In a real world scenario, users may need to configure different number of acceptors vs threadless processes.  Currently, there is a threadless process dedicated for each acceptor process.

1. Add a `--num-acceptors` flag
2. Decouple threadless pool setup/teardown from acceptor pool class

**Describe alternatives you've considered**

**Additional context**
This might be necessary if you are not expecting too many concurrent client requests but each request is a heavy request (e.g. file upload / file download).

abhinavsingh:
Addition of `--num-acceptors` flag will spring a surprise for several folks using `proxy.py` with a custom `--num-workers` flag.  One option is to send out a PR to all known forks using a custom `--num-workers` flag :).

Another option, that is being implemented currently is:
- If `--num-workers` flag is used w/o `--num-acceptors` flag, then default value of `num_acceptors = num_workers`
- For all other scenarios, use the auto-default value or value provided via flags.
- Forks directly using `proxy.TestCase` will not be impacted by this PR because, `--num-acceptors 1` is also hardcoded in `TestCase` now.
- Fixes #706 

Above rules will prohibit users from expecting the following scenario:
- Use `--num-workers` flag only and expect `proxy.py` to spawn auto-default value number of `acceptors`.
- But this is OK, because most of the times users won't need more acceptors, but more workers depending upon type of payload/requests.  You only need several acceptors if you are expecting a tsunami of requests :)

NOTE: Also renames `proxy.pool` to `proxy.acceptors`


webknjaz:
@abhinavsingh this branch you have flooded the GHA backlog: https://github.com/abhinavsingh/proxy.py/actions?query=is%3Aqueued.
This is because your PR branch lives in the upstream repository.
So each time you push to this branch, there are two events that fire in this repo: `push` and `pull_request`. Each of them starts a workflow.

One of the solutions I've seen in the wild is to forbid creating branches in the upstream repo (we actually have a bot that tries to prevent this at work, in ansible/ansible). And then, the feature branches (except for `devel` and release branches) should exist in the forks of people (even for the maintainers).
Not only this helps with the CI, but it also prevents flooding the contributors' forks with garbage branches. Whenever somebody forks a repo, they get a copy of all the branches at that point in time. The upstream branches get updated/removed but the forks don't get those updates and they remain polluted with unimportant branches. Also, since the forks happen at different points in the repo history, different users/forks would get those branches in different states which may be confusing, especially for Git beginners. This would also spread the load and allow the contributors to get some CI results in their forks before creating a PR even. This is because their pushes to the feature branch will create a `push` event in their fork (never affecting the upstream CI in any way) and the upstream PRs would generate `pull_request` events which is fine even with many updates since there's auto-cancelation.

Action items:
1) Cancel the stale workflows, they get auto-canceled for PRs but not for the regular in-repo pushes.
2) Consider maybe moving the repo into a dedicated org and changing your workflow a little bit.

abhinavsingh:
> this branch you have flooded the GHA backlog: https://github.com/abhinavsingh/proxy.py/actions?query=is%3Aqueued.

Lollll, yes, totally.  Apologies for that.  Last night I was onto something, I knew several tests would break, some may even hang, and I didn't cared to fix the tests until I could get it working right.  Out of habit, I kept pushing code upstream too, resulting into hung workflows.

Indeed I did notice duplicate workflows being triggered, one each for `push` and `pull_request`.  I think, one option is to disable workflows for `push` (w/o an open PR).  Because currently, even if I don't create a PR, workflows are being triggered for every push.  Wdyt?  Will it also impact workflows for pushes into an active PR.

> One of the solutions I've seen in the wild is to forbid creating branches in the upstream repo (we actually have a bot that tries to prevent this at work, in ansible/ansible). And then, the feature branches (except for `devel` and release branches) should exist in the forks of people (even for the maintainers).

I see what you mean here.  IIUC, TL;DR is that development should happen in forks which will avoid running workflows within the main repo context.  Though, I think duplicate workflows will still get triggered for every push into personal forks.

I am thinking, may be right solution is to disable `push` only workflows until there is an associated PR for that branch.  This will avoid running unintended workflows both for the upstream repo and personal forked repos.  Of-course disabling on `push` will prohibit developers from seeing CI results until they make a PR.

To make sure we are on the same ground, here is my understanding of how workflows are getting triggered:

1. Configured for `push` and `pull_request` actions
2. Triggered for every `push` (w/o a PR)
3. Triggered when a `pull_request` is created
4. Triggered when there is a `push` into an active PR.  This `push` then in turn also triggers (duplicate) corresponding `pull_request` workflows.

Please correct me if wrong.  I would like to tackle `Item-4` where duplicate workflows are getting triggered.

> Not only this helps with the CI, but it also prevents flooding the contributors' forks with garbage branches. Whenever somebody forks a repo, they get a copy of all the branches at that point in time. The upstream branches get updated/removed but the forks don't get those updates and they remain polluted with unimportant branches.

You have a valid point here.

> This would also spread the load and allow the contributors to get some CI results in their forks before creating a PR even.

Makes sense, we expect forks to have a valid workflows setup.

> 1. Cancel the stale workflows, they get auto-canceled for PRs but not for the regular in-repo pushes.

Done

> 2\. Consider maybe moving the repo into a dedicated org and changing your workflow a little bit.

Totally, soon we'll put `proxy.py` under @jaxl-innovations-private-limited organization.  I have done repo moves under org in the past, but still wondering, any side-effects that we must be concerned about before making this move?

One action item would be to update all `abhinavsingh/` references to `jaxl-innovations-private-limited/` in the codebase artifacts.

webknjaz:
> > this branch you have flooded the GHA backlog: [abhinavsingh/proxy.py/actions?query=is%3Aqueued](https://github.com/abhinavsingh/proxy.py/actions?query=is%3Aqueued).
> 
> Lollll, yes, totally. Apologies for that. Last night I was onto something, I knew several tests would break, some may even hang, and I didn't cared to fix the tests until I could get it working right. Out of habit, I kept pushing code upstream too, resulting into hung workflows.

I don't see pushing to the remote as inherently problematic. The only problem is that you don't have a separate fork with separate rate limits in this case so you end up influencing others.

> Indeed I did notice duplicate workflows being triggered, one each for `push` and `pull_request`. I think, one option is to disable workflows for `push` (w/o an open PR). Because currently, even if I don't create a PR, workflows are being triggered for every push. Wdyt? Will it also impact workflows for pushes into an active PR.

I don't like this because it'll also prevent feature branches from being tested in forks when there's no PR.

> > One of the solutions I've seen in the wild is to forbid creating branches in the upstream repo (we actually have a bot that tries to prevent this at work, in ansible/ansible). And then, the feature branches (except for `devel` and release branches) should exist in the forks of people (even for the maintainers).
> 
> I see what you mean here. IIUC, TL;DR is that development should happen in forks which will avoid running workflows within the main repo context. Though, I think duplicate workflows will still get triggered for every push into personal forks.

Not exactly. For forks, the `push` event within the GitHub platform will happen in the context of those forks. When there's a PR to the upstream, from a fork, on PR branch updates the fork will get `push` events and the upstream will get `pull_request` events, `push` will never trigger workflow runs in upstream in this case (but it will trigger those in the corresponding fork repo; the workflows in the fork will not show up among checks in PRs or anywhere in upstream for that matter).

> I am thinking, may be right solution is to disable `push` only workflows until there is an associated PR for that branch. This will avoid running unintended workflows both for the upstream repo and personal forked repos. Of-course disabling on `push` will prohibit developers from seeing CI results until they make a PR.

This is exactly what I dislike. I sometimes push to my fork to see the results before making a PR, with this change, you'd steal this ability and force posting a PR that may not yet be ready (not even for Draft).

> To make sure we are on the same ground, here is my understanding of how workflows are getting triggered:
> 
> 1. Configured for `push` and `pull_request` actions
> 2. Triggered for every `push` (w/o a PR)
> 3. Triggered when a `pull_request` is created
> 4. Triggered when there is a `push` into an active PR.  This `push` then in turn also triggers (duplicate) corresponding `pull_request` workflows.

The `push` **event** doesn't trigger the `pull_request` event. They appear independently as a result of `git push`. The duplication is only visible when a given PR is coming from the same repository and is not a concern in other cases.

> > This would also spread the load and allow the contributors to get some CI results in their forks before creating a PR even.
> 
> Makes sense, we expect forks to have a valid workflows setup.

With your proposal to disable the `push` trigger, you'd break this.

> > 2. Consider maybe moving the repo into a dedicated org and changing your workflow a little bit.
> 
> Totally, soon we'll put `proxy.py` under @jaxl-innovations-private-limited organization.

Is this org tightly bound to proxy.py? If not, maybe create a separate one?

> I have done repo moves under org in the past, but still wondering, any side-effects that we must be concerned about before making this move?

No, GH will create redirects. Both on HTTP level and also for Git too.

> One action item would be to update all `abhinavsingh/` references to `jaxl-innovations-private-limited/` in the codebase artifacts.

This can be done later since the links won't be broken due to redirects.

webknjaz:
> No, GH will create redirects. Both on HTTP level and also for Git too.

One thing I remembered is that post-migration, you'd need to make sure that the same GitHub Apps are installed there (because the installations are bound to GH profiles and orgs).

PR #715, opened 2021-11-09
Switch versioning to be SCM-based

webknjaz:


webknjaz:
This depends on the presence of Git tags.

webknjaz:
@abhinavsingh as a reminder — tag the appropriate commit for the last release you've done on `develop` and then re-run the workflows here to see the version updated.

Makefile
```diff
@@ -2,7 +2,8 @@ SHELL := /bin/bash
 
 NS ?= abhinavsingh
 IMAGE_NAME ?= proxy.py
-VERSION ?= v$(shell python -m proxy --version)
+#VERSION ?= v$(shell bash -c "python -m setuptools_scm --version \| awk '{print$3}'")
```
abhinavsingh, line 5: I think we want to use this one once upstream tag has been set?

I added a `v2.3.1` tag and now:

```
python -m setuptools_scm --version | awk '{print$3}' 
2.3.2.dev140+g094e30d
```

```
python -m setuptools_scm --version | awk '{print $$3}' | sed 's/\+/--/'
Version
```

abhinavsingh: Aah, I noticed double `$$` used due to escape reasons.

- Any specific reason why you uses `--` and not `-`?  Within makefile the `VERSION` value currently resolves to `v2.3.2.dev140--g094e30d.d20211113`
- Also how do we go from `2.4` or `3.0`, may be answer is in `scmtools` flag?

abhinavsingh (commented):
I'll need to catch up with `scmtools` default to make more sense out of these PR.  But all looks good from the top.  Please see my comments.

I am yet to check the follow-up PR.  You may have addressed this already there.  But another question that comes to mind, how do we remove `.dev*` suffix for stable releases, so that `PyPi` release is `vX.Y.Z` instead of `vX.Y.Z.dev`

Makefile
webknjaz: I guess I forgot to drop the commented-out line...

webknjaz: > Aah, I noticed double `$$` used due to escape reasons.

Yeah, it took me a while to fight this problem. Make has syntax that is too complex and is not really necessary for Python projects. It'd be great to migrate to `tox` entirely but that's a separate topic.

> * Any specific reason why you uses `--` and not `-`?  Within makefile the `VERSION` value currently resolves to `v2.3.2.dev140--g094e30d.d20211113`

No reason, I just needed to replace that plus and thought that maybe a single dash may be ambiguous. FWIW I can replace it with one dash, it won't affect anything.

> * Also how do we go from `2.4` or `3.0`, may be answer is in `scmtools` flag?

By tagging the commit.

webknjaz:
> I am yet to check the follow-up PR. You may have addressed this already there.

That PR focuses on adding a trigger for cutting releases from GH UI and connecting that to jobs that would perform actual publishing to PyPI. It's incomplete IIRC. I'll undraft it once this one is merged.

> But another question that comes to mind, how do we remove `.dev*` suffix for stable releases, so that `PyPi` release is `vX.Y.Z` instead of `vX.Y.Z.dev`

That suffix is only added to non-tagged commits. If you tag the current commit, that tag will be used exactly as it is.

PR #735, opened 2021-11-14
`--enable-proxy-protocol` : HAProxy Protocol v1

abhinavsingh:
Implements v1 spec as defined here https://www.haproxy.org/download/1.8/doc/proxy-protocol.txt

Additionally, added a setup step (`make lib-scm-version`), required now to work with source code locally out of repo.


Makefile
```diff
@@ -89,6 +88,9 @@ lib-dep:
 		-r requirements-release.txt \
 		-r requirements-tunnel.txt
 
+lib-scm-version:
+	@echo "version = '$(VERSION)'" > proxy/common/_scm_version.py
```
webknjaz, line 92: This misses out on the `tuple_version`

tests/http/test_proxy_protocol.py
```diff
@@ -0,0 +1,86 @@
...
+
+    :copyright: (c) 2013-present by Abhinav Singh and contributors.
+    :license: BSD, see LICENSE for more details.
+"""
+import unittest
+
+from proxy.http.parser import ProxyProtocol, PROXY_PROTOCOL_V2_SIGNATURE
+
+
+class TestProxyProtocol(unittest.TestCase):
```
webknjaz, line 16: Did you know that inheriting from `unittest.TestCase` prevents you from using many powerful pytest features?

abhinavsingh: Totally, lot of legacy finger memory :).  Lots of enhancements that we must make in these respect.  E.g. total `pytest` adoption.  May be upgrade linting from `autopep8` to `black`.  Hopefully some day soon :D

Makefile
abhinavsingh: Can you expand, per `_scm_version.pyi`

```
# This stub file is necessary because `_scm_version.py`
# autogenerated on build and absent on mypy checks time
version: str
```

So I simply dumped this version string in there.  It seems to work.  Are we missing something here?

tests/http/test_proxy_protocol.py
webknjaz: Personally, I can't stand black because it doesn't let you make important decisions. But it's fine for newbies who don't know how to write code consistently. I find most of the autoformatters harmful in many because in some cases they silently hide legitimate problems that a machine cannot solve because of the lack of human context. This is why I opt-in for very specific formatting cases that are unlikely to shadow real problems.

PR #736, opened 2021-11-14
Reuse a pre-existing `version_tuple` from SCM

webknjaz:
This should've been done from the start.


proxy/common/_scm_version.pyi
```diff
@@ -1,3 +1,6 @@
 # This stub file is necessary because `_scm_version.py`
 # autogenerated on build and absent on mypy checks time
+from typing import Tuple, Union
+
 version: str
+version_tuple: Tuple[Union[int, str], ...]
```
abhinavsingh, line 6: Aah I see,`version_tuple` is in here.

abhinavsingh: Should we also update `Makefile` target too?

webknjaz: You could. I'm not familiar with GNU/make too much so I won't be exploring this. Plus, after agreeing to drop it, it probably won't make much sense trying to solve it there.

abhinavsingh: I have double thoughts about `Makefile` and `master` branch.  I'll add my thoughts here for makefile

So I removed makefile locally and started using `tox` based workflow locally.  After a few days, I still feel makefile should be retained and I miss it :).  It helps provide the following *(we may or may not use it within GHA workflows)*:
  - A simple shortcut to various targets across the project
  - Everything cannot be powered by `tox` e.g. non-python bits in the project
  - As a user, make file gives me autocompletion.  So, I don't have to depend on my shell history.  With `tox` I am totally dependent on learning the `tox` syntax and flags.  Which might be a good thing to do, but should not be necessary.

If we choose to not use makefile in workflow, we'll need to keep the `lib-` bits in sync.  But even that is OK, frankly, `lib-` targets will very rarely change.  Wdyt?

proxy/common/_scm_version.pyi
```diff
@@ -1,3 +1,6 @@
 # This stub file is necessary because `_scm_version.py`
 # autogenerated on build and absent on mypy checks time
+from typing import Tuple, Union
+
 version: str
```
abhinavsingh, line 5: I see what you meant by tuple bit earlier.  It worked within Makefile target then because it simply resulted in creation of `_scm_version.py` file :)

proxy/common/_scm_version.pyi
webknjaz: @abhinavsingh

> * A simple shortcut to various targets across the project

This sounds valid. OTOH, personally I invoke tox by finding it among the previous commands in the terminal with `Ctrl+R` (I also have `fzf` integrated for fast fuzzy search) so I don't really experience the need to have shortcuts.

What exactly are you missing? Tox maintains its own venvs but you can instruct it to create one for you outside of its normal structure with the `--devenv` option if you want to wrap that. But then, you can only work with one env at a time so you'd probably need to have some way to switch the desired toxenv for development, I guess.

> * Everything cannot be powered by `tox` e.g. non-python bits in the project

This is not really true. You can run arbitrary commands you like. The experience is a bit different but it is totally possible. You'll probably just need `skip_install = true` for that.

> * As a user, makefile gives me autocompletion.  So, I don't have to depend on my shell history.  With `tox` I am totally dependent on learning the `tox` syntax and flags.  Which might be a good thing to do, but should not be necessary.

I have autocompletion with tox. It's configured via an oh-my-zsh plugin. Other shell probably can do this similarly.

> If we choose to not use makefile in workflow, we'll need to keep the `lib-` bits in sync. But even that is OK, frankly, `lib-` targets will very rarely change. Wdyt?

You could do that but you need to keep the source of truth is tox since it's what can be mapped to the CI in a flexible maner.

PR #737, opened 2021-11-14
Make Git archives `setuptools-scm` compatible

webknjaz:
This makes it possible to `pip install` from Git archives created from tags (like the GitHub ones).


webknjaz:
@abhinavsingh you can merge this right away. The check job won't run because of the file-based filtering.

abhinavsingh:
> @abhinavsingh you can merge this right away. The check job won't run because of the file-based filtering.

If changes to these files must run a check, may be we can include these within lib workflow.  Wdyt?

For now let's go ahead.

webknjaz:
Maybe. OTOH they are very unlikely to change again.

PR #740, opened 2021-11-15
Handle invalid/malformed data from clients in HttpParser

### LINKED ISSUE #127: [HttpParser] Handle invalid/malformed request chunks from client
**Describe the bug**
Currently, `HttpParser` works well assuming incoming client is not notorious :)  Handle cases when client can send invalid chunks in requests.

**To Reproduce**
1. Run `proxy.py` as `proxy.py --pac-file proxy.pac`
2. `telnet localhost 8899` and hit `Enter` key to send line feed
```
$ telnet localhost 8899
Trying ::1...
Connected to localhost.
Escape character is '^]'.

Connection closed by foreign host.
```
3. See error

```
2019-10-11 11:12:06,277 - pid:55396 [E] run:2215 - Exception while handling connection <socket.socket fd=27, family=AddressFamily.AF_INET6, type=SocketKind.SOCK_STREAM, proto=0, laddr=('::1', 8899, 0, 0), raddr=('::1', 49902, 0, 0)>
  File "./proxy.py", line 2205, in run
    teardown = self.run_once()
  File "./proxy.py", line 2173, in run_once
    teardown = self.handle_events(readables, writables)
  File "./proxy.py", line 2138, in handle_events
    teardown = self.handle_readables(readables)
  File "./proxy.py", line 2076, in handle_readables
    self.request.parse(client_data)
  File "./proxy.py", line 615, in parse
    more, raw = self.process(raw)
  File "./proxy.py", line 625, in process
    self.process_line(line)
  File "./proxy.py", line 667, in process_line
    self.url = urlparse.urlsplit(line[1])
IndexError: list index out of range
2019-10-11 11:12:06,279 - pid:55396 [I] access_log:1955 - ::1:49902 -  None
```

**Expected behavior**
`HttpParser` must either ignore invalid chunks or be strict about it and drop the connection.

JerryKwan:
1. Add HTTP request line validation in `proxy/http/parser/parser.py`
2. Add test case for the modification
3. Fixes #127


proxy/http/parser/parser.py
```diff
@@ -301,16 +301,18 @@ def _process_line_and_headers(self, raw: bytes) -> Tuple[bool, bytes]:
...
                 # We expect to receive entire proxy protocol v1 line
                 # in one network read and don't expect partial packets
                 self.protocol.parse(raw)
             else:
                 line = raw.split(WHITESPACE)
-                self.method = line[0].upper()
-                self.set_url(line[1])
-                self.version = line[2]
-                self.state = httpParserStates.LINE_RCVD
+                if len(line) == 3:
```
abhinavsingh, line 311: Thinking more about it.  I feel may be we must just raise an exception when line doesn't have sufficient parts as per the specification.  Reason why I think raising an exception is better than silently ignoring invalid packets:

1. Issue #127 is generally not an issue for a well behaved clients
2. This is however a possible attack vector because malicious users/hackers can start sending specially crafted request to the server.
3. If we choose to ignore invalid bytes, a hacker can still overwhelm the server by streaming invalid chunks to the server.
4. E.g. imagine, someone by mistake connects a SSH client on port `8899`.  What will happen then?  Will our parser keep on rejecting invalid bytes, expecting to see a valid HTTP packet someday?   May be answer is to drop such invalid clients as soon as possible.

I think we can simply raise a `NotImplementedError` for now which will eventually result in dropping the client connection from the core.   Wdyt?

tests/http/test_http_parser.py
```diff
@@ -23,6 +23,13 @@ class TestHttpParser(unittest.TestCase):
     def setUp(self) -> None:
         self.parser = HttpParser(httpParserTypes.REQUEST_PARSER)
 
+    def test_issue_127(self) -> None:
+        self.parser.parse(CRLF)
+        self.assertEqual(self.parser.state, httpParserStates.INITIALIZED)
+        raw = b'qwqrqw!@!#@!#ad adfad\r\n'
```
abhinavsingh, line 29: Thank you for this test case.  This prompted me to post previous comment i.e. raise exception instead of ignoring.

Because if in this test we add the following:

```python
while 1:
    self.parser.parse(b'qwqrqw!@!#@!#ad adfad\r\n')
```

server will keep on spinning (because it will ignore invalid bytes), resulting in lower performance for other clients connecting to the server

abhinavsingh (commented):
Thank you!!! PTAL at my comments.

proxy/http/parser/parser.py
JerryKwan: That's right, we should quick fail other than waiting for valid input

proxy/http/parser/parser.py
```diff
@@ -300,19 +299,28 @@ def _process_line_and_headers(self, raw: bytes) -> Tuple[bool, bytes]:
...
+            # http://www.iana.org/assignments/http-methods/http-methods.xhtml
             if self.protocol is not None and self.protocol.version is None:
                 # We expect to receive entire proxy protocol v1 line
                 # in one network read and don't expect partial packets
                 self.protocol.parse(raw)
             else:
                 line = raw.split(WHITESPACE)
-                if len(line) == 3:
-                    self.method = line[0].upper()
+                if len(line) == 3 and line[0] in httpMethods:
```
abhinavsingh, line 316: Thank you.  But additional check on `line[0] in httpMethods` must be removed.  Reason being, `HttpParser` is common for web and proxy server components.  While this check is valid for a proxy server, a web server can use custom HTTP method verbs, which are valid as per specification.

If a parsed method verb is invalid, then it should not be a parser error, but actual error response must be returned from server handlers. E.g. returning `Method not supported error`.  For reference see screenshot below from linked RFC

Screen Shot 2021-11-15 at 7 33 00 PM

proxy/http/parser/parser.py
```diff
@@ -300,19 +299,28 @@ def _process_line_and_headers(self, raw: bytes) -> Tuple[bool, bytes]:
...
+                    self.method = line[0]
                     self.set_url(line[1])
                     self.version = line[2]
                     self.state = httpParserStates.LINE_RCVD
+                else:
+                    # raise exception
+                    raise HttpProtocolException("Invalid request line")
```
abhinavsingh: +1 `HttpProtocolException` is better than `NotImplementedError` as parser is within the `http` sub-module.

tests/http/test_http_parser.py
```diff
@@ -17,18 +17,22 @@
 from proxy.http import httpStatusCodes, httpMethods
 from proxy.http.parser import HttpParser, httpParserTypes, httpParserStates
```
abhinavsingh, line 19: No need of an extra line.  Idea is to keep all sub-module level imports together.

abhinavsingh (commented):
PTAL

proxy/http/parser/parser.py
```diff
@@ -300,19 +299,28 @@ def _process_line_and_headers(self, raw: bytes) -> Tuple[bool, bytes]:
...
             if self.protocol is not None and self.protocol.version is None:
                 # We expect to receive entire proxy protocol v1 line
                 # in one network read and don't expect partial packets
                 self.protocol.parse(raw)
             else:
                 line = raw.split(WHITESPACE)
-                if len(line) == 3:
-                    self.method = line[0].upper()
+                if len(line) == 3 and line[0] in httpMethods:
+                    self.method = line[0]
```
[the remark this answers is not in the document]
JerryKwan: As the RFC said, the method is case-sensitive, that's why I removed the upper().

proxy/http/parser/parser.py
JerryKwan: Maybe we should update the supported method according to 
https://www.iana.org/assignments/http-methods/http-methods.xhtml

tests/http/test_http_parser.py
JerryKwan: Agree, Will remove the extra line

proxy/http/parser/parser.py
JerryKwan: I will remove the method check for now

proxy/http/parser/parser.py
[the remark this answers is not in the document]
abhinavsingh: Awesome, good catch.

proxy/http/parser/parser.py
abhinavsingh: Yes lets add those methods within `httpMethods` named tuple.  About restricting verbs to only defined verb, I am not sure about it still.  E.g. why stop someone from using a custom verb in first place.  How does it hurt.  In the end the server will anyways return unsupported method if received verb is invalid within its context.  Hence I think we must allow parsing custom verbs and let server handlers do their job or rejecting the request.

Wdyt?

JerryKwan: We can remove the method restriction for now and let server handlers to do there job, that will give more flexibility

proxy/http/parser/parser.py
```diff
@@ -300,17 +299,28 @@ def _process_line_and_headers(self, raw: bytes) -> Tuple[bool, bytes]:
         return len(raw) > 0, raw
 
     def _process_line(self, raw: bytes) -> None:
+        # defer import to solve the circular import problem
+        from ..exception import HttpProtocolException
```
abhinavsingh, line 303: For now let's just throw 'NotImplementedError` and get past this PR.  We can always clean-up in a follow up PR, because we are running into circular import exception.  If fixing circular import is not straightforward, let's not try to fight two things in a single PR.  Leave a `TODO: Runs into circular import error when using HttpProtocolException` and possibly link to this comment.  Thank you!!!

abhinavsingh (commented):
PTAL

proxy/http/parser/parser.py
JerryKwan: Got it.  Have updated the codes accordingly, thank you

tests/http/test_http_parser.py
```diff
@@ -16,13 +16,23 @@
...
 class TestHttpParser(unittest.TestCase):
 
     def setUp(self) -> None:
         self.parser = HttpParser(httpParserTypes.REQUEST_PARSER)
 
+    def test_issue_127(self) -> None:
+        with self.assertRaises(HttpProtocolException):
+            self.parser.parse(CRLF)
+
+        with self.assertRaises(HttpProtocolException):
```
abhinavsingh, line 31: You forgot to change exception type in test.

abhinavsingh (commented):
PTAL

tests/http/test_http_parser.py
JerryKwan: Sorry, It's my fault. Fixed already.

abhinavsingh:
I highly recommend you to setup https://github.com/abhinavsingh/proxy.py#setup-local-environment or at-least run those commands manually before pushing to upstream.  To avoid unnecessary to and fro.  Example now we ran into this issue:

Screen Shot 2021-11-18 at 6 38 13 AM

at-least make sure to run

`make lib-mypy && make lib-flake8` to ensure linting is fine.  You may still run into other issues.  For total satisfaction, run `make` to run all tests locally.  :)

I know it's a PITA but is necessary to maintain code quality over time.

JerryKwan:
Good advice. 
I an using a Windows development environment right now, I will change to a Linux environment and setup the hooks later

abhinavsingh (approved):
LGTM

PR #743, opened 2021-11-15
Fix dist version in CI/CD on pushes to `develop`

webknjaz:


abhinavsingh:
Yay!!!! Yes.  I am wondering why was `test.pypi` release skipped for latest PR merge.  https://github.com/abhinavsingh/proxy.py/actions/runs/1463441959

May be because of same `.dev` suffix?

webknjaz:
> Yay!!!! Yes. I am wondering why was `test.pypi` release skipped for latest PR merge. https://github.com/abhinavsingh/proxy.py/actions/runs/1463441959
> 
> May be because of same `.dev` suffix?

It's not a push to develop, it's a PR build:
> Triggered via pull request

abhinavsingh:
> It's not a push to develop, it's a PR build:

Somehow when I copy the link of workflow from repo home-page, it gave me that URL of a PR.

Better to just check it here -- https://github.com/abhinavsingh/proxy.py/commits/develop -- See last PR merge into dev and associated workflow.

webknjaz:
I think you're confusing the events and check reports here. You did not merge that PR into devel. You have fast-forwarded the PR branch to devel, it's different.

GitHub's Checks API operates on commits. Commits may belong to different branches. When you updates a PR, it's HEAD started pointing at the same commit as the devel's HEAD. This caused a `pull_request` `synchronize` event to report a Check that shows up on that commit regardless on which branch you're looking at. Branches are just floating labels/pointers by themselves, you cannot say that the check was reported for a branch but for a commit that happened to be the one that branch pointed at, at certain point in time.

When you first created the commit https://github.com/abhinavsingh/proxy.py/commit/79f51e716a6dd0ffdc4755dbd2dab8b08bbaf902, this is when the `push` event happened. When multiple Checks are reported on the same commit, they are all (or most of them are) displayed for that commit but the view you're looking at may confuse you. Only when you go to a certain Check's details, you can discover its workflow run and from there, see what event triggered it.

To find what you're really looking for, you can go to the repo-global Actions tab and then, select the workflow and apply a few filters (event + branch). This is how you can reliably find that workflow publishing to TestPyPI: https://github.com/abhinavsingh/proxy.py/actions/workflows/test-library.yml?query=event%3Apush+branch%3Adevelop.

abhinavsingh:
Thank you for the insight, makes more sense now.  Indeed, I was confusing things between `push` and `pull request` events.

PR #751, opened 2021-11-18
Add all IANA assigned HTTP method names

abhinavsingh:
Additionally:
- Fixes the additional `v` prefix in printed version
- Now every `make` invocation regenerates `proxy/common/_scm_version.py` for latest version string


.gitignore
```diff
@@ -26,4 +26,6 @@ cover
 htmlcov
 dist
 build
+
 proxy/public
+docs/_build
```
webknjaz, line 31: Why did you do this? It seems like you're duplicating tox's effort. It's a bad idea.

Makefile
```diff
@@ -82,14 +83,13 @@ lib-clean:
...
 	pip install \
 		-r requirements.txt \
 		-r requirements-testing.txt \
 		-r requirements-release.txt \
-		-r requirements-tunnel.txt
-
-lib-scm-version:
-	@echo "version = '$(VERSION)'" > proxy/common/_scm_version.py
+		-r requirements-tunnel.txt \
+		-r docs/requirements.txt
```
webknjaz, line 92: It's a bad idea to mix up test and docs requirements. They should be separate.

Makefile
```diff
@@ -114,6 +114,16 @@ lib-release-test: lib-package
 lib-release: lib-package
 	twine upload dist/*
 
+lib-doc:
+	pushd docs && \
+	python -m sphinx \
```
webknjaz, line 119: No need to maintain a separate copy of the build process. Call `tox` like everybody else in the ecosystem. The docs deps are conceptually different from the test ones. By mixing them up you increase the chance of hitting pointless dependency conflicts.
Just do `tox -e build-docs`.

tests/http/test_http_parser.py
```diff
@@ -24,10 +24,10 @@ def setUp(self) -> None:
         self.parser = HttpParser(httpParserTypes.REQUEST_PARSER)
 
     def test_issue_127(self) -> None:
-        with self.assertRaises(NotImplementedError):
+        with self.assertRaises(ValueError):
```
webknjaz, line 27: These CMs like `pytest.raises()`/`pytext.warns()` usually accept the second argument for the message. It is the best practice to always pass that for precise exception expectations.

write-scm-version.sh
```diff
@@ -0,0 +1,46 @@
+#!/bin/bash
+#
+# write-scm-version.sh exists because `proxy.py``
+# auto-detects it's next version from git.  Hence,
+# for `proxy.py` to work `proxy/common/_scm_version.py`
+# file must be auto-generated with necessary information.
+#
+# For CI/CD, this file is generated via `tox` integration.
+# For local development (without editable install), you
```
webknjaz, line 9: I'd argue, that for local development, an editable install must be used which would take care of this w/o the need for this script.

.gitignore
abhinavsingh: I used `python -m sphinx` and I ended up with `docs/_build`.  May be we should update that to use `tox`

Makefile
abhinavsingh: Agreed.  This should be part of `make lib-doc` itself

tests/http/test_http_parser.py
abhinavsingh: Agreed, this needs to be covered across all `raise` expectations I guess.

write-scm-version.sh
abhinavsingh: Agreed.  I'll most likely remove this bootstrapping once readme is updated with editable install as go-to developer setup step.

.gitignore
webknjaz: Well, the idea is that you don't do this because tox already has that wrapped and managed under `.tox/`.

Makefile
```diff
@@ -114,6 +114,16 @@ lib-release-test: lib-package
...
+		--keep-going \
+		-b dirhtml \
+		-d _build/doctrees \
+		-D language=en . _build/html && \
+	popd && \
+	open docs/_build/html/index.html
```
webknjaz, line 125: Also, `open` only exists on macOS. Here you exclude all the Linux-based contributors who'd need an `xdg-open.

Makefile
webknjaz: https://github.com/abhinavsingh/proxy.py/pull/752

Makefile
abhinavsingh: Thank you, will consider presence of `xdg-open` or add a OS check

PR #752, opened 2021-11-18
Fix Makefile to use tox for docs

webknjaz:


Makefile
```diff
@@ -115,14 +114,8 @@ lib-release: lib-package
...
-	pushd docs && \
-	python -m sphinx \
-		--keep-going \
-		-b dirhtml \
-		-d _build/doctrees \
-		-D language=en . _build/html && \
-	popd && \
-	open docs/_build/html/index.html
+	python -m tox -e lint && \
+	xdg-open .tox/build-docs/docs_out/index.html
```
abhinavsingh, line 118: This will break for `mac`.  I think we can define a top-level OPEN variable which captures what binary to use.  E.g. `xdg-open`, `open` or something else in future.

abhinavsingh (commented):
PTAL

Makefile
webknjaz: I don't know how to do this in the makefile but feel free to commit to this branch.

abhinavsingh: How about something like this:

```
OPEN=$(shell which open)
UNAME := $(shell uname)
ifeq ($(UNAME), Linux)
OPEN=$(shell which xdg-open)
endif
```

* Put these lines at the top of the Makefile before `.PHONY` section.  It uses `open` by default and `xdg-open` on Linux.
* Replace all `open` with `$(OPEN)`

webknjaz: That seems out of the scope. Commit, if you want.

abhinavsingh: I am not sure how to push to your branch :).  I can clone your PR locally and try it out, but not push to it.  To push to your branch, I'll have to add your repo as upstream locally or fork on Github I guess, correct?  Though I am not sure if this may require some permission from your end.  Don't know, I haven't pushed to a PR before.  Lemme know :)

webknjaz: GitHub has a read-only namespace. Here's how you can synchronize it:
```
git config --add remote.origin.fetch '+refs/pull/*/head:refs/pull/origin/*'
git fetch --all
```
(given that you point `origin` to this repo)

You cannot push there but you'd get the code locally.
To push you'll indeed need to add a remote for my work:
```
git add webknjaz git@github.com:webknjaz/proxy.py.git
git fetch --all
```
Alternatively, you can even push by the remote URL w/o adding it with a name:
```
git push git@github.com:webknjaz/proxy.py.git <local_branch>:<remote_fork_branch>
```

This is all possible with the `Allow edits by maintainers` checkbox on PRs.

But you can also just go to the Files tab in the PR, click on the edit button (:pencil2:), right on the GH UI. And commit from there: https://github.com/webknjaz/proxy.py/edit/maintenance/sphinx-make/Makefile?pr=%2Fabhinavsingh%2Fproxy.py%2Fpull%2F752.

abhinavsingh: Thank you, that opens up new possibilities.  I know about remote origin setup but last time when I tried it, I received a permission error.  Will check back on it tonight.

webknjaz: AFAIK, GitHub also has their Git wrapper CLI called `gh` that encapsulates some of this but I haven't used it personally.

PR #756, opened 2021-11-19
[IntegrationTest] Use `127.0.0.1` as target address and a random port

abhinavsingh:
1. Using `localhost` may not work reliably depending upon underlying IPv6 / IPv4 resolution differences.  Since we start proxy on `127.0.0.1`, let's use the same target address
2. Random port is necessary to avoid port conflicts during parallel test execution.

Also fixes doc spellchecks CI failing due to missing words

### DIFF
--- a/docs/spelling_wordlist.txt
+++ b/docs/spelling_wordlist.txt
@@ -5,3 +5,9 @@ scm
 Threadless
 threadless
 youtube
+socio
+sexualized
+https
+www
+html
+faq
--- a/tests/integration/test_integration.py
+++ b/tests/integration/test_integration.py
@@ -5,7 +5,11 @@ from typing import Generator
 
 import pytest
 
+from proxy.common.utils import get_available_port
 from proxy.common._compat import IS_WINDOWS  # noqa: WPS436
+
+
+PROXY_PY_PORT = get_available_port()
 
 
 # FIXME: Ignore is necessary for as long as pytest hasn't figured out
@@ -22,6 +26,7 @@ def _proxy_py_instance() -> Generator[None, None, None]:
     proxy_cmd = (
         'proxy',
         '--hostname', '127.0.0.1',
+        '--port', str(PROXY_PY_PORT),
         '--enable-web-server',
     )
     proxy_proc = Popen(proxy_cmd)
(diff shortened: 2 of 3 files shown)


docs/spelling_wordlist.txt
```diff
@@ -5,3 +5,9 @@ scm
 Threadless
 threadless
 youtube
+socio
```
webknjaz, line 8: Maybe add these to the document and not globally?

docs/spelling_wordlist.txt
```diff
@@ -5,3 +5,9 @@ scm
...
+sexualized
+https
+www
+html
```
webknjaz, line 12: This will probably be unnecessary if you fix the document to have an explicit link by wrapping that link with angled brackets.

tests/integration/test_integration.py
```diff
@@ -5,9 +5,13 @@
 
 import pytest
 
+from proxy.common.utils import get_available_port
 from proxy.common._compat import IS_WINDOWS  # noqa: WPS436
 
 
+PROXY_PY_PORT = get_available_port()
```
webknjaz, line 12: It's best to use an ephemeral port. Otherwise, it's still possible to hit race conditions.

abhinavsingh: Yep I would have, but situation here is a little different.  Proxy is not in embedded mode bu started from command line.  So we must then either parse port from logs or let proxy write the port somewhere or use pid file to infer the port  :).  I went for the easy solution for now.

docs/spelling_wordlist.txt
abhinavsingh: Totally, I added these because CI was failing and making me nervous, haha.  Let me address them properly as a follow up PR.  I have a few bits more related to doc.  Will pull in as part of that.

abhinavsingh:
Merging this for now.  I'll address doc lint related stuff in a follow-up doc PR.  Updating logo and adding doc link in README.md.   Last night #753 became excessively big so I gave up on it, will move parts of it as separate PRs.  The license fix will still require editing almost every single file.  Will repurpose #753 for the same.

PR #769, opened 2021-11-21
Async `get_events`, `handle_event`, `handle_readables`, `handle_writables`

abhinavsingh:
## Async

This is another step closer to everything async.  In this PR, we have updated the core to use async callbacks.  This currently doesn't impact the plugins which remain sync.  In future though, plugin interface will be upgraded to use async.  This should not break the plugins but will require an update i.e. add `async` keyword to methods you are using from the plugin interface.

## Performance

This update also speeds up our async execution by about 50%.  Now we don't wait on all work to finish, but instead we bail out on first work completion (or timeout), giving chance to process new incoming work.

## Local executor flag

Further, added a `--local-executor` flag.  This is one step closer to bringing executors within acceptor context for CPU affinity.  This will avoid usage of `send_handle` and `recv_handle` which are slow as per profiler result.  Executor framework itself will remain, which can be independently used outside of the acceptor context.   Currently, `--local-executor=False` is default, which means each acceptor will continue to delegate work to multi-core executor pool.

## Fd not Sock

Now all callbacks work with `fileno` instead of `sock` objects.  This makes maintenance easy within the core when tracking active `fds`.  Also results in less overhead.


.flake8
```diff
@@ -65,6 +65,7 @@ extend-ignore =
   N802    # FIXME: func name should be lowercase
   N816    # FIXME: mixed case var name
   P101    # FIXME: format string with unindexed params
+  PT007   # FIXME: wrong values type in @pytest.mark.parametrize, expected tuple
```
abhinavsingh, line 68: @webknjaz Had to add this, otherwise `flake8` was reporting error.  I think it's a false positive.  I am using `parametrize` with `indirect=True`.  Here is an example usage:

```python
@pytest.mark.parametrize(
        '_setUp',
        [('test_modify_post_data_plugin')],
        indirect=True,
    )   # type: ignore[misc]
```

docs/conf.py
```diff
@@ -274,4 +274,5 @@
     (_py_class_role, 'UUID'),
     (_py_class_role, 'WebsocketFrame'),
     (_py_class_role, 'Url'),
+    (_py_class_role, '_asyncio.Task'),
```
abhinavsingh, line 277: @webknjaz This is kind of annoying, specially because there is no hint from the editor itself.  Until one run the CI, they won't discover such nitty things, resulting in slow iteration, discouragement from contributions.   E.g. if a person need to make 3-turns to just fix CI, it's simply discouraging.

Either we should look into some sort of Editor configuration (I already checked in my `.vscode`), or relax such warnings which are simply unnecessary.  Evaluating by weight, value that this adds is not worth it to slow down contributions and iterations.

PS: Problem is locally entire flow cannot be run, specially because of `pylint`.  There are other issues I observed e.g. stale content in `.tox`, `doc` where I had to delete these folders to get going.  I will list them down separately.

proxy/common/backports.py
```diff
@@ -0,0 +1,82 @@
...
+    Adopted from https://wiki.python.org/moin/PythonDecoratorLibrary#Cached_Properties
+    © 2011 Christopher Arndt, MIT License.
+
+    NOTE: We need this function only because Python in-built are only available
+    for 3.8+.  Hence, we must get rid of this function once proxy.py no longer
+    support version older than 3.8.
+
+    .. spelling::
+
+       backports
```
abhinavsingh, line 53: It's unclear why `sphinx` want us to tell it spelling of the filename itself.  We have files named `acceptor`, `executor`, `backport` but still we need to declare them as spellings?  Another redundancy IMHO.

abhinavsingh: PS:  Again this is not known to the authors until they run the `spellcheck-doc` flow.  Nothing from the editor.  I am not sure such `sphinx` extensions exist for VSCode.

proxy/core/acceptor/local.py
```diff
@@ -0,0 +1,131 @@
...
+import socket
+import logging
+import argparse
+import selectors
+import threading
+import multiprocessing.synchronize
+
+from multiprocessing import connection
+from typing import Optional, Tuple, List, Dict
+from typing import Any      # noqa: W0611   pylint: disable=unused-import
```
abhinavsingh, line 26: Such things were again annoying.  I had to disable this for `flake8` and `pylint` separately :(.  Another redundancy imho.  We should not check for common errors across `flake8` and `pylint`, specially if it disrupts the smooth development flow.

requirements-testing.txt
```diff
@@ -5,6 +5,8 @@ flake8==4.0.1
 pytest==6.2.5
 pytest-cov==3.0.0
 pytest-xdist == 2.4.0
+pytest-mock==3.6.1
```
abhinavsingh, line 8: I ran into another scenario where I had to add `pytest-mock==3.6.1` within `tox.ini` lint section.  Can't we avoid doing it?  Unsure why we are not installing `requirements-testing` in there, which would have got this in automatically.  Had to go through CI to figure this out.

tests/http/exceptions/test_http_proxy_auth_failed.py
```diff
@@ -8,43 +8,44 @@
...
-    @mock.patch('selectors.DefaultSelector')
-    @mock.patch('socket.fromfd')
-    def setUp(
-        self,
-        mock_fromfd: mock.Mock,
-        mock_selector: mock.Mock,
-    ) -> None:
-        self.mock_fromfd = mock_fromfd
-        self.mock_selector = mock_selector
+    @pytest.fixture(autouse=True)   # type: ignore[misc]
```
abhinavsingh, line 27: One good thing, we are now on `pytest` for the most part :)

.flake8
webknjaz: No, it's not a false-positive. It is set up to expect a tuple for the param variants and you use a list in your example. It's a style I chose for consistency.
```diff
 @pytest.mark.parametrize(
         '_setUp',
-        [('test_modify_post_data_plugin')],
+        (
             ('test_modify_post_data_plugin'),
         ),
         indirect=True,
     )   # type: ignore[misc]
```

So drop the ignore and follow the selected style:
```suggestion
```
You could adjust the settings of this plugin later if you want.

tests/plugin/test_http_proxy_plugins.py
```diff
@@ -24,21 +25,23 @@
...
-    @mock.patch('socket.fromfd')
-    def setUp(
-        self,
-        mock_fromfd: mock.Mock,
-        mock_selector: mock.Mock,
-    ) -> None:
+class TestHttpProxyPluginExamples(Assertions):
+
+    @pytest.fixture(autouse=True)   # type: ignore[misc]
+    def _setUp(self, request: Any, mocker: MockerFixture) -> None:
```
webknjaz, line 38: It's unnatural to use Java-style in the context of Python / pytest. The previous name was a byproduct of the original `unittest` module in stdlib "copied" from Java. But nowadays, unless you're forced to use `unittest` directly, don't use camelCase for methods/functions.

tests/plugin/test_http_proxy_plugins.py
```diff
@@ -24,21 +25,23 @@
...
+        self.mock_fromfd = mocker.patch('socket.fromfd')
```
webknjaz, line 39: Instead of trying to have side effects with unclear influence targets on the class instance, make a fixture that constructs and returns this. In fact, make a cascade of pure-function based fixtures that are easy to compose and redefine at multiple levels.

tests/test_assertions.py
```diff
@@ -0,0 +1,26 @@
...
+
+    :copyright: (c) 2013-present by Abhinav Singh and contributors.
+    :license: BSD, see LICENSE for more details.
+"""
+from typing import Any
+
+
+class Assertions:
+
+    def assertTrue(self, obj: Any) -> None:
```
webknjaz, line 16: Why would you do something like this? It's quite unnatural for pytest.

tests/plugin/test_http_proxy_plugins_with_tls_interception.py
```diff
@@ -178,32 +176,59 @@ def send(raw: bytes) -> int:
         )
         self.assertFalse(self.protocol_handler.work.has_buffer())
 
-    def test_modify_post_data_plugin(self) -> None:
+        #
```
webknjaz, line 179: ?

requirements-testing.txt
webknjaz: Because the venv is cached and tox is not always able to figure out if the deps changed. You can add `-r` to tox when you want to force it to recreate a venv based on the new deps. Or you could `rm -rf .tox` (or some of its venvs) if you want.

webknjaz: Oh wait, you're asking about `lint`. I guess that's because flake8 doesn't really need all of the deps listed here. OTOH we could distill a subset of the deps in a separate file that could indeed be reused.
The current dependencies setup is quite a mess so I skipped solving this problem for now. Ideally, a pip-tools managed structure should be implemented for proper separation of concerns.

tests/http/exceptions/test_http_proxy_auth_failed.py
webknjaz: It's still far from following pytest's common usage patterns.

README.md
```diff
@@ -2175,3 +2211,5 @@ https://github.com/abhinavsingh/proxy.py/issues/new
 - Single threaded server.
 
 For detailed changelog refer to release PRs or commit history.
+ history.
+ history.
```
webknjaz, line 2215:
```suggestion
```

check.py
```diff
@@ -53,16 +53,18 @@
                 sys.exit(1)
 
 # Update README.md flags section to match current library --help output
```
webknjaz, line 55: If you move the output to a sphinx document, you could just use https://sphinxcontrib-programoutput.rtfd.io and https://sphinxcontrib-autoprogram.rtfd.io to stop rehardcoding this chunk.

docs/conf.py
webknjaz: This is because of the autogenerated API docs. You cannot do anything about this. These are generated from annotations that sometimes point at things that are not documented in the stdlib docs. You could disable turning warnings into errors entirely and allow committing broken docs for which you won't have any signals for whether everything is correct or not.

proxy/common/_compat.py
```diff
@@ -1,9 +1,19 @@
-"""Compatibility code for using Proxy.py across various versions of Python.
+# -*- coding: utf-8 -*-
+"""
+    proxy.py
```
webknjaz, line 3: Any reason to have unnecessary indentation in every docstring?

proxy/common/backports.py
webknjaz: This is because apidoc generates RST files for each module now. In those files, the document titles contain importable names. And sometimes those names are not in the dictionary. You could somehow modify the templates to wrap the titles with double backticks, I guess. I haven't attempted to solve this because I've spent a lot of effort already and I'm simply exhausted at this point (having a lot more projects to maintain). Sometimes I figure out better ways and then go and update many projects but not for this problem, currently.

As for the editor, it's uncommon to run sphinx from it (or any other command for that matter). Moreover, nobody is supposed to run it directly because you'll end up with people running it with different configurations you don't know anything about and complaining about things that work differently in CI and RTD. Just add a command for running tox if you want.

Also, spellcheck is third-party. You can ask here https://github.com/sphinx-contrib/spelling.

P.S. As I mentioned earlier, many projects just document their API manually, without autogenerating pages for all modules.

proxy/common/backports.py
```diff
@@ -0,0 +1,82 @@
...
+
+    :copyright: (c) 2013-present by Abhinav Singh and contributors.
+    :license: BSD, see LICENSE for more details.
+"""
+import time
+
+from typing import Any
+
+
+class cached_property:
```
webknjaz, line 16: I don't think you should vendor this. It's more sustainable to just use a common pattern of falling back to use a backport: https://docs.python.org/3/library/functools.html#functools.cached_property + https://pypi.org/p/backports.cached-property.

proxy/common/_compat.py
```diff
@@ -1,9 +1,19 @@
-"""Compatibility code for using Proxy.py across various versions of Python.
+# -*- coding: utf-8 -*-
```
webknjaz, line 1: This is only necessary if you support Python 2.
```suggestion
```

proxy/common/_compat.py
```diff
@@ -1,9 +1,19 @@
...
+"""
+    proxy.py
+    ~~~~~~~~
+    ⚡⚡⚡ Fast, Lightweight, Pluggable, TLS interception capable proxy server focused on
+    Network monitoring, controls & Application development, testing, debugging.
+
+    :copyright: (c) 2013-present by Abhinav Singh and contributors.
+    :license: BSD, see LICENSE for more details.
+
+    Compatibility code for using Proxy.py across various versions of Python.
```
webknjaz, line 11: With this change, you're breaking the PEP 257 style compatibility.

proxy/core/acceptor/local.py
webknjaz: It's hard to do this because many of the checks are not 100% equivalent.

proxy/core/acceptor/local.py
```diff
@@ -0,0 +1,131 @@
...
+        self.executor_locks = executor_locks
+        self.event_queue = event_queue
+        # Incremented every time work() is called
+        self._total: int = 0
+        self._selector: Optional[selectors.DefaultSelector] = None
+        self._works: Dict[int, Work] = {}
+
+    def run_once(self) -> bool:
+        try:
+            payload = self.evq.get(block=True, timeout=0.1)
```
webknjaz, line 69: You should only put this line in try/except. It's an antipattern to have more than one instruction in such blocks.

proxy/core/acceptor/local.py
```diff
@@ -0,0 +1,131 @@
...
+            if isinstance(payload, bool) and payload is False:
+                return True
+            assert isinstance(payload, tuple)
+            conn, addr = payload
+            addr = None if addr == '' else addr
+            self.dispatch(conn, addr)
+        except queue.Empty:
+            pass
```
webknjaz, line 77: For the pattern of suppression, use `with contextlib.suppress(queue.Empty):`

tests/plugin/test_http_proxy_plugins_with_tls_interception.py
```diff
@@ -178,32 +176,59 @@ def send(raw: bytes) -> int:
...
+    @pytest.mark.asyncio    # type: ignore[misc]
+    @pytest.mark.parametrize(
+        '_setUp',
+        [('test_man_in_the_middle_plugin')],
+        indirect=True,
+    )   # type: ignore[misc]
+    async def test_man_in_the_middle_plugin(self) -> None:
+        await self.protocol_handler._run_once()
+
+        self.assertEqual(self.mock_sign_csr.call_count, 1)
```
webknjaz, line 216: Having the actual assertion in a different place will probably make the pytest output worse (it patches the bytecode of the `assert` statements to retrieve more context about the local variables in the place where it's present)

tests/integration/test_integration.py
```diff
@@ -1,4 +1,15 @@
...
+"""
+    proxy.py
+    ~~~~~~~~
+    ⚡⚡⚡ Fast, Lightweight, Pluggable, TLS interception capable proxy server focused on
+    Network monitoring, controls & Application development, testing, debugging.
+
+    :copyright: (c) 2013-present by Abhinav Singh and contributors.
+    :license: BSD, see LICENSE for more details.
+
+    Test the simplest proxy use scenario for smoke.
```
webknjaz, line 11: Per PEP 257, a docstring should start with a title up to 72 chars followed by an empty line and then, the rest of the prose.

tests/http/exceptions/test_http_proxy_auth_failed.py
abhinavsingh: Do you mean without classes?  Anything wrong about `autouse=True` here within class context.

tests/plugin/test_http_proxy_plugins.py
abhinavsingh: Will address these changes separately as this residual effect is spread across all the test files.

setup.cfg
```diff
@@ -88,7 +87,7 @@ keywords =
   Python3
 
 [options]
-python_requires = >= 3.6
```
webknjaz, line 91: Could you keep the compatibility until the next release, at least? aiohttp v3.8 uses proxy.py in tests and still runs the tests against Python 3.6 on that branch. We won't be able to use proxy.py with this change. And the previous release still causes a critical dependency tree problem. Or maybe you could release the code before this PR with py36 and then drop it?

tests/plugin/test_http_proxy_plugins.py
abhinavsingh: Totally, I see what you mean.  Ideally fixtures should have returned the result back to be used within test.  But at this point, I didn't want to disrupt the entire test suite.  I had to use `pytest` because I needed `asyncio` support, as core was updated to use `async` almost everywhere.

I'll come back to tests and nail them properly.  With `pytest` we can also reduce a lot of redundant code in tests.  When I revisited them today, I could see the redundancy :).  But left it for later.

proxy/http/handler.py
```diff
@@ -374,27 +364,30 @@ def _selected_events(self) -> SelectedEventsGeneratorType:
...
+        return (events, readables, writables)
 
-    def _run_once(self) -> bool:
-        with self._selected_events() as (readables, writables):
-            teardown = self.handle_events(readables, writables)
+    async def _run_once(self) -> bool:
+        events, readables, writables = await self._selected_events()
+        try:
+            teardown = await self.handle_events(readables, writables)
             if teardown:
```
webknjaz, line 373: This should be in the `else:` part of try-except. Besides, it could be simplified to a one-liner `return bool(teardown)`.

tests/test_assertions.py
abhinavsingh: Will deprecate this eventually, I didn't want to add it, but I also wanted to keep my focus on async changes rather than removal of all `self.assert*` methods.  Will revisit all this mess again

tests/plugin/test_http_proxy_plugins_with_tls_interception.py
abhinavsingh: Thank you, yes a residual.  I had to copy lines above `#` out of `setUp` because I couldn't make async fixture work correctly.  So I added a marker for later revisit, could have been a TODO indeed.

.flake8
abhinavsingh: Thank you, lemme address this right in.

README.md
abhinavsingh: Thank you.  I think `check.py` is faulty and inserting this automagically.  Lemme address it.

proxy/core/base/tcp_server.py
```diff
@@ -52,19 +52,23 @@ def __init__(self, *args: Any, **kwargs: Any) -> None:
         self.must_flush_before_shutdown = False
         if self.flags.unix_socket_path:     # pragma: no cover
             logger.debug(
-                'Connection accepted from {0}'.format(self.work.address),
+                'Work#{0} accepted from {1}'.format(
+                    self.work.connection.fileno(), self.work.address,
+                ),
             )
```
webknjaz, line 58: There are a few problems here:
1) You could use an f-string
2) It's best to use named placeholders, using numbers hurts readability because it increases the cognitive complexity (the reader would need to match numbers with the format args and count their positions each time)
3) It's an antipattern to pre-render log entries. For any use of `logging`, you should pass a string with placeholders and then, separate values. There are a few reasons for this: (1) the logging framework may be deciding not to output this log entry depending on the configuration on a completely different level so rendering it is wasteful; (2) some tools that collect stats on what logging invocations are hit use the log line to group the data and it works best with a template line being the same and not different every time the control flow hits this place.
```suggestion
            logger.debug(
                'Work#%d accepted from %s',
                self.work.connection.fileno(),
                self.work.address,
            )
```

check.py
abhinavsingh: Wow nice.  I wanted something like this now that we have sphinx.  Totally, we can automate flags section of `README.md` too.

I was looking for a tool like this to automate example snippets within the doc-string.  My idea about docs is a little different than what you proposed above.  I think docs should be part of the code and not outside of it.   When writing docs, I want to include example snippets within class doc-string.  But problem is, those snippets may soon run out of sync with the actual code.  So all snippets must be within a directory, say `snippets` or we can re-use `examples` folder for it.  Then, I want to simply add something like this in the class doc-string

```
.. snippet::

   examples/snippets/acceptors.py#L10-L13
```

and this should render into actual code snippet either embedded via GitHub APIs or embedded during doc generation. 
 We can write tests for snippets too.  This will ensure that doc readers never run into surprises because I missed updating the doc somewhere :)

Unsure if linked plugins can do this out-of-box.  I'll dig into them.

proxy/common/_compat.py
abhinavsingh: Not really.  I started fixing these in one of the branches, but will need to revisit later again.

proxy/common/_compat.py
abhinavsingh: Totally, I am keeping it for consistency right now.  This is being addressed as part of doc-string fix.

proxy/common/_compat.py
abhinavsingh: Yep :(. I think will be addressed with doc-string fix.  `Compatibility code for using Proxy.py across various versions of Python.` will then be the only content of doc-string

tests/plugin/test_http_proxy_plugins_with_tls_interception.py
abhinavsingh: I agree, it's an anti-pattern within pytest eco-system.  Lemme deprecate `Assertions` too within this PR :D

tests/integration/test_integration.py
abhinavsingh: Totally.  I'll keep these changes and address them as part of doc-string.  Easier to fix via `awk`/`sed` in a separate branch :) Actually `Makefile` was broken for sometime.  I fixed it today which started running `check.py` and it caught all files without library header.  Had to add back in missing files.

setup.cfg
abhinavsingh: Totally.  Won't submit this immediately.  I fear something somewhere will break and make folks unhappy.

Good idea.  Let's make a release and then push this in.

Btw, about release, I earlier restricted `PYPI_TOKEN` for master branch.  Because I wanted to keep the same workflow i.e. release should happen from master branch.  I haven't yet looked deeply into `pypi-release` workflow steps.  Might need to address if they are dependent on `develop` somehow.

So here is what I'll do:

1) PR `v2.4.0` from develop to master
2) Tag `v2.4.0` pointing to master
3) Our PyPi release trigger.

One idea can be, we make a `rc` release first, or may be `a`, `b` releases.  Because this will be the first time we run the release workflow.  Wdyt?

abhinavsingh: Let me also spend more time thinking about it.  I read about a few thing in `asyncio`, but need to try them out.  I feared I might break 3.7+, so was being lazy about it.  But lemme double check if we can maintain 3.6 support.

proxy/core/acceptor/local.py
abhinavsingh: Cool this is new for me.  Now it looks much cleaner to read *(will push later)*

```python
def run_once(self) -> bool:
    with contextlib.suppress(queue.Empty):
        payload = self.evq.get(block=True, timeout=0.1)
        if isinstance(payload, bool) and payload is False:
            return True
        assert isinstance(payload, tuple)
        conn, addr = payload
        self.dispatch(conn, addr)
    return False
```

proxy/http/handler.py
abhinavsingh: Good catch, since `handle_events` are supposed to return `teardown` value itself, can simply do `return self.....`

setup.cfg
abhinavsingh: Also note that, usage of `asyncio.run` is in part of the code which is only necessary for `--threaded` execution.  Background, `_run_once` used to be sync, and now it has become async.  Due to which when execution is started in a thread, `run` loop is unable to call async `_run_once`, for which now we use `asyncio.run`.

This possibly could be resolved by starting a new loop for every thread.  PS:  Threaded execution performance is already going down due to unnecessary async overhead.  Adding it's own loop for every request handling in a thread, will likely bring down the performance further down.  When I say down, I mean comparable to the `--threadless` execution.

I was thinking of sharing the event loop with threads, but I read it's unsafe and can lead to unintended behavior.  Still I think we can resolve it *(possibly)*.  Will dig into it over this week.

.gitignore
```diff
@@ -8,7 +8,6 @@
 .mypy_cache
 .hypothesis
 .tox
-.python-version
```
webknjaz, line 11: @abhinavsingh do not remove this. It's used by pyenv for saving the user-selected dir-local envs/pythons and mustn't be committed.

.gitignore
```diff
@@ -8,7 +8,6 @@
 .mypy_cache
 .hypothesis
 .tox
```
webknjaz, line 10:
```suggestion
.tox
.python-version
```

check.py
webknjaz: Including the program output is a separate task and should be implemented via those extensions.

As for the snippet inclusion, there's a few points.
1) doctests exist and they are testable. It's a well-known standard, use it. https://docs.python.org/3/library/doctest.html / https://www.sphinx-doc.org/en/master/usage/extensions/doctest.html. The runner for them is already added to the CI so just add those.
2) It is possible to include parts of files as code blocks in RST/Sphinx. You can use https://www.sphinx-doc.org/en/master/usage/restructuredtext/directives.html#directive-literalinclude which also allows you to include the contents partially and add some highlighting. Use this in the actual documents.
3) Do not use `.. literalinclude` in the docstrings. It's unreadable, untestable, and unmaintainable. It's a very bad idea.

You can include huge bits of docs in docstrings but many projects have moved away from this practice because a lot of prose may make it harder to deal with actually coding stuff. Plus you'll probably be unable to create https://diataxis.fr structured site (which you should). This is why these approaches are most often combined.

docs/conf.py
```diff
@@ -1,5 +1,5 @@
 # pylint: disable=invalid-name
-# Requires Python 3.6+
+# Requires Python 3.7+
```
webknjaz, line 2: This is not true. The code here does not use any features that are Python 3.7 specific. And it does not have exactly the same dependencies as the runtime. Sphinx is being executed in a separate env that mustn't be tightly coupled with the test/dev/runtime envs.

proxy/core/acceptor/acceptor.py
```diff
@@ -117,8 +134,15 @@ def run_once(self) -> None:
                         if mask & selectors.EVENT_READ:
                             if self.sock is not None:
                                 conn, addr = self.sock.accept()
-                                addr = None if addr == '' else addr
-                                self._work(conn, addr)
+                                logging.debug(
```
webknjaz, line 137: The same comment as in https://github.com/abhinavsingh/proxy.py/pull/769#discussion_r753842787 applies here (and in any similar places).

proxy/core/acceptor/acceptor.py
```diff
@@ -117,8 +134,15 @@ def run_once(self) -> None:
...
+                                    'Accepting new work#{0}'.format(conn.fileno()),
+                                )
+                                if self.flags.local_executor:
+                                    assert self._local
+                                    self._local.evq.put((conn, addr))
+                                else:
+                                    addr = None if addr == '' else addr
+                                    self._work(conn, addr)
```
webknjaz, line 145: this could probably be
```suggestion
                                    self._work(conn, addr or None)
```

proxy/core/acceptor/local.py
webknjaz: Almost there: it's still wrong to expect `isinstance()` or `self.dispatch()` to raise `queue.Empty`.

proxy/core/acceptor/local.py
```diff
@@ -0,0 +1,131 @@
...
+        return False
+
+    def run(self) -> None:
+        self._selector = selectors.DefaultSelector()
+        try:
+            while True:
+                if self.run_once():
```
webknjaz, line 84: Why don't you wrap just this line with the suppression? This way you'd be able to dedent a whole lot of code. With the lower level of nesting, it's easier to reason about the code (readability/maintainability).

proxy/core/acceptor/local.py
webknjaz: Also, you could just put it under an if-block as per common usage pattern
```suggestion
from typing import TYPE_CHECKING
if TYPE_CHECKING:
    from typing import Any
```
and you won't have these linting violations. The linters are correct: you do not currently refer this name in runtime.

proxy/core/acceptor/threadless.py
```diff
@@ -93,44 +97,35 @@ def selected_events(self) -> Generator[
...
+        assert len(tasks) > 0
+        finished, self.unfinished = await asyncio.wait(
+            tasks,
+            timeout=(self.flags.num_workers / 2) *
+            DEFAULT_SELECTOR_SELECT_TIMEOUT,
+            return_when=asyncio.FIRST_COMPLETED,
+        )
+        for task in finished:
+            if task.result():
+                self.cleanup(task._work_id)     # type: ignore
```
webknjaz, line 127: Instead of ignoring all possible type errors, put the exact error code that you're seeing to prevent accidentally shadowing more problems that may arise in the future:
```suggestion
                self.cleanup(task._work_id)     # type: ignore[a-specific-error]
```

proxy/core/acceptor/threadless.py
```diff
@@ -178,13 +173,38 @@ def cleanup(self, work_id: int) -> None:
...
+            tasks.add(task)
+        return tasks
+
+    async def run_once(self) -> None:
         assert self.loop is not None
-        with self.selected_events() as (readables, writables):
-            if len(readables) == 0 and len(writables) == 0:
+        events, work_by_ids = await self._selected_events()
+        try:
+            if len(work_by_ids) == 0:
```
webknjaz, line 199: I think the common idiom would be
```suggestion
            if not work_by_ids:
```

proxy/core/acceptor/threadless.py
```diff
@@ -178,13 +173,38 @@ def cleanup(self, work_id: int) -> None:
...
                 # Remove and shutdown inactive connections
                 self.cleanup_inactive()
```
webknjaz, line 201: have you considered only wrapping this instruction with try/except?

tests/http/exceptions/test_http_proxy_auth_failed.py
webknjaz: It's just something flexible. Plus yes, w/o classes you could also do autouse if really needed.

setup.cfg
webknjaz: I suppose you could replace `asyncio.run()` with `loop.run_until_complete()`. As for starting many loops, it's most definitely is a bad idea. FWIW, if it's hard for you to bring back 3.6, let's just release the current latest commit on `develop` as 2.4.0 (well, or its beta/rc maybe).

webknjaz: > Btw, about release, I earlier restricted `PYPI_TOKEN` for master branch. Because I wanted to keep the same workflow i.e. release should happen from master branch. I haven't yet looked deeply into `pypi-release` workflow steps. Might need to address if they are dependent on `develop` somehow.

I don't think you need to restrict it to a specific branch. The idea is that you can point at any commit sha when triggering the workflow, if that's what you want.

> So here is what I'll do:
> 
> 1. PR `v2.4.0` from develop to master
> 2. Tag `v2.4.0` pointing to master
> 3. Our PyPi release trigger.

That's incorrect. The idea is that you shouldn't have to do multiple steps manually. Because of the human factor, you may make mistakes in a complex multi-step process.
Instead, it's currently set up to be a single-step trigger with approvals along the way. You only need to trigger the workflow and the tagging should happen automatically, after the publishing to the PyPI.

> One idea can be, we make a `rc` release first, or may be `a`, `b` releases. Because this will be the first time we run the release workflow. Wdyt?

Alright. Here's what we can do. Just go to https://github.com/abhinavsingh/proxy.py/actions/workflows/test-library.yml, trigger it, only inputting `2.4.0a0` in the `release-version` field. Wait until it gets to the normal PyPI job, you'll see an approval request on the workflow run page and maybe in your inbox. Confirm it.
Don't worry about releasing to regular PyPI. Since it's an alpha version, people will not upgrade to it by default + you'll be able to `yank` the release on PyPI right away (if you want, that is).
Do this only to confirm that the release pipeline is set up properly. No other reason. If this works, we'll proceed with trying to merge dev to master, and then trying to release its last commit with `2.4.0a1` but explicitly filling out the `release-commitish` field too. Once you're satisfied with the results, we can think about `2.4.0rc0` or going straight to `2.4.0`. But that last decision is to be made later, don't worry. Just do the above to be able to practice the process _in a safe way_.

.gitignore
abhinavsingh: Are we adding back `.python-version`?  I removed it because I deleted my local copy of `.python-version`.  It was causing issue in `zsh` when switching to `Python 3.6`.

.gitignore
abhinavsingh: Ok makes sense, just saw this comment.

proxy/core/acceptor/local.py
abhinavsingh: Is there a `--fix` equivalent for these which will automatically add the suggestion instead of throwing errors?

proxy/core/acceptor/threadless.py
abhinavsingh: Refactored code and this line is now within `Threadless`, applied there manually.

abhinavsingh: `# type: ignore[attr-defined]` works

proxy/core/acceptor/threadless.py
abhinavsingh: I am not sure about it.  `work_by_ids` is a dictionary, so idea here is to check if it contains any key.  I'll covert these to list/tuple and might be able to use this construct then.

proxy/core/acceptor/threadless.py
abhinavsingh: After further refactoring try/except no longer used here

proxy/core/base/tcp_server.py
abhinavsingh: Applied manually, GH says file was deleted.  I moved them around

proxy/core/acceptor/local.py
abhinavsingh: This didn't work, ran into runtime errors:

```console
File "/Users/abhinavsingh/Dev/proxy.py/proxy/core/acceptor/local.py", line 32, in LocalExecutor
    def __init__(self, *args: Any, **kwargs: Any) -> None:
NameError: name 'Any' is not defined. Did you mean: 'any'?
```

docs/conf.py
abhinavsingh: But why do we have to add our own lib classes here e.g. `WebsocketFrame` and such

proxy/common/backports.py
abhinavsingh: > I haven't attempted to solve this because I've spent a lot of effort already and I'm simply exhausted at this point (having a lot more projects to maintain)

I totally agree.  You have totally transformed CI/CD/Packaging stuff for `proxy.py`.  Very happy about it.  Really appreciate all the contributions.

> P.S. As I mentioned earlier, many projects just document their API manually, without autogenerating pages for all modules.

But then docs can run out of sync with code and it's a double work to maintain code and keep docs up-to date.  Wdyt about it?

requirements-testing.txt
abhinavsingh: > OTOH we could distill a subset of the deps in a separate file that could indeed be reused.

Makes sense

> The current dependencies setup is quite a mess so I skipped solving this problem for now.

Totally.  I was thinking do we still need `requirements-* and simply not put them within `tox.ini`.  What will be the side-effects?

abhinavsingh: > Ideally, a pip-tools managed structure should be implemented for proper separation of concerns.

Will look into it, haven't used it yet.

tests/plugin/test_http_proxy_plugins_with_tls_interception.py
abhinavsingh: Created https://github.com/abhinavsingh/proxy.py/issues/775 to address various `pytest` related migration changes.  Will address this `#` once as part of `fixtures` setup.

tests/integration/test_integration.py
abhinavsingh: Created https://github.com/abhinavsingh/proxy.py/issues/776 for `PEP 257` related stuff

proxy/common/_compat.py
abhinavsingh: Tracking here https://github.com/abhinavsingh/proxy.py/issues/776

proxy/common/_compat.py
abhinavsingh: Tracking here https://github.com/abhinavsingh/proxy.py/issues/776

tests/plugin/test_http_proxy_plugins.py
abhinavsingh: Tracking here https://github.com/abhinavsingh/proxy.py/issues/775

.gitignore
webknjaz: Yes. It's what pyenv users commonly use. I don't know what your problem was but it was probably specific to your (mis)configuration or maybe you just used it wrong. Anyway, even if you stop using pyenv, you shouldn't prevent the contributors from using it.

requirements-testing.txt
webknjaz: I think it should be a separate effort, this PR is messy enough.

proxy/core/acceptor/local.py
webknjaz: > Is there a `--fix` equivalent for these which will automatically add the suggestion instead of throwing errors?

No, linters exist to point out problems. They don't have a human context to make decisions on how you want to address the problems you've created.

webknjaz: > This didn't work, ran into runtime errors:
> 
> ```
> File "/Users/abhinavsingh/Dev/proxy.py/proxy/core/acceptor/local.py", line 32, in LocalExecutor
>     def __init__(self, *args: Any, **kwargs: Any) -> None:
> NameError: name 'Any' is not defined. Did you mean: 'any'?
> ```

This means that you've actually started using the variable and don't need a conditional import. Now you need it to be unconditional and there's no violation you had earlier.

docs/conf.py
webknjaz: It depends on the exact case. I can't guess a universal answer. For the cases where it's coming from auto-generated pages, it may be that I excluded a few modules because they were causing unsolvable problems. Another case would be when you manually try referring to this object but use a wrong/incomplete reference. Show me where it happens and I'll tell you why exactly.

webknjaz: Oh, you're probably referring to the one I added originally. I was ignoring a lot of things because it was easier to get the initial setup merged this way. Those should be fixed. Most of the objects are documented and should be referred to via their full names. As in `package.subpackage.module.WebsocketFrame`. But if you just use `WebsocketFrame`, it's hard for Sphinx to guess what you mean. There's syntax for making sure that when you use the full reference, it'll still be rendered as the short one. It RST, it needs a tilde:
```rst
:class:`~package.subpackage.module.WebsocketFrame`
```
the above will be rendered as `WebsocketFrame` in HTML while referencing a specific documented object.
If you do
```rst
:class:`package.subpackage.module.WebsocketFrame`
```
it will be rendered as `package.subpackage.module.WebsocketFrame`.
You can also enter a custom anchor:
```rst
:class:`some thing <package.subpackage.module.WebsocketFrame>`
```
This one will be rendered as `some thing` while linking the proper target.

proxy/common/backports.py
webknjaz: > > P.S. As I mentioned earlier, many projects just document their API manually, without autogenerating pages for all modules.
> 
> But then docs can run out of sync with code and it's a double work to maintain code and keep docs up-to date. Wdyt about it?

I need to clarify that with the hybrid approach people often use some of the sphinx directives (`autofunction`?) that may extract parts of the content from docstrings while having accompanying prose in the documents. I think you need to try creating more documents and follow https://diataxis.fr just to have a feeling of how this all works. When you experiment with these things more, maybe it'll become easier to judge where to use prose and what to leave fully autogenerated. Per my observations, the manually crafted parts are usually ones that describe the interconnection between things and the public API while the internal objects/reference may be more generated. I think that in terms of Diataxis, it's reasonable to have separately-maintained Tutorials and How-To Guides while Explanations and References may get more of the generated stuff. Still, it's a balance that you need to find empirically.

requirements-testing.txt
webknjaz: > Totally. I was thinking do we still need `requirements-*` and simply not put them within `tox.ini`. What will be the side-effects?

Missed this part. Don't do this. I know that it's tempting but each tox env has a single list of deps while different runtime envs (OS/Python) would need separate lists of pins. Plus no tools are able to auto-update `tox.ini` anyway.
Oh, and once there's lockfiles/constaints with the exact pins, it'll be unsustainable to maintain hundreds of autogenerated lines in the middle of the tox config.

proxy/core/acceptor/local.py
abhinavsingh: Yep `Any` is part of most constructor.

docs/conf.py
abhinavsingh: Cool, so one step if to remove word references and include correct syntax for class references.  We should be able to deprecate most of these then.

proxy/core/acceptor/threadless.py
webknjaz: All empty containers are falsy. `bool({})` / `bool([])` / `bool(())` / `bool(set())` are all evaluated to `False`. This is why when you see `if a_dict_var:`, it's semantics is "check whether this dict is empty".

proxy/core/acceptor/local.py
webknjaz: (not that it's a good idea, though)

docs/conf.py
abhinavsingh: Example, I removed `--threadless`.  I realize why you had to add it then to get the integration going.  Should be possible to avoid that entire list of any and py roles.

abhinavsingh: Tracking in https://github.com/abhinavsingh/proxy.py/issues/776

proxy/common/backports.py
abhinavsingh: Cool.  I need to investigate more into this before finalizing on an approach.

My inspiration is coming from Flutter/Dart land, where code contains entire doc site as doc strings.  So as a developer, I hardly have to leave the editor and go to some doc site for reference.  Of-course, flutter team also maintains dedicated tutorial, manuals and showcases for the front page of doc site.

abhinavsingh: Tracking here https://github.com/abhinavsingh/proxy.py/issues/776

proxy/common/backports.py
abhinavsingh: I thought about it.  Instead of a `pip install` I'll prefer a baked in implementation.  I picked this up from some StackOverflow thread.  But point being, we can further customize and build on top of this base implementation for `proxy.py` needs.  I don't think we'll be limited to a simple `@cached_property`, we'll likely need caching strategies and such in future.  Then, we'll able to modify this to our taste.

tests/http/exceptions/test_http_proxy_auth_failed.py
abhinavsingh: Tracking here https://github.com/abhinavsingh/proxy.py/issues/775

tests/test_assertions.py
abhinavsingh: Tracking here https://github.com/abhinavsingh/proxy.py/issues/775

Thought I'll get this in, but let's keep for now.  Might address soon anyways.

abhinavsingh:
@webknjaz Python 3.6 is back.  Might have to apply similar fix for Python 3.7.  I see tests hanging *(most likely on e2e integration tests)* for 3.7 too.  Amazingly 3.7 on Windows works :).   Python is sometimes funny

proxy/common/backports.py
webknjaz: You still need to have enough information in the docstrings, of course. The developers will be able to call `help()` on objects in the default Python REPL, or `?` in IPython, and various editor integration will be able to extract those. You could also configure Sphinx to generate a man page that would be usable by the downstream consumers so regular Linux distro users would also be able to get that via manpages (`man proxy-py`, for example). But that all doesn't replace the purpose of the types of the docs. I really recommend following the Diataxis principles. Ubuntu is about to adopt it right now and you can also chat with its creator in a dedicated channel on the Write The Docs Slack instance. I've heard that our tech writers at Red Hat have a similar set of guidelines they follow.

proxy/common/backports.py
webknjaz: Vendoring things may be undesired from a legal PoV. In that case you'd have to comply with the license requirements which often demand that the original code is distinguishable from that you add. This is why vendored libs usually live in their own directory, in separate files/subfolders, and you'll have to maintain their licenses+legal notices in prominent places. Plus, adding your modifications on top would need to be clean, probably separate, and/or well-documented.
Also, `pip install` already gets all the deps and is a de-facto standard.
Vendoring implies added maintenance burden: you'll have to manually figure out if the upstream project has any security vulnerabilities, apply fixes and re-release your project quickly (because otherwise, you steal the ability to get security fixes from your users). This is non-automatable, especially not with pieces of code that is not distributed in a standard way. You could solve this by integrating the tool that pip itself uses (and only because it technically cannot rely on the deps for self-bootstapping): https://pypi.org/p/vendoring.
This should be meant as the last resort and all the projects using it eventually move away from this technique whenever possible.
At least two projects I'm involved with that used to rely on vendored deps have moved away from them when it became possible (CherryPy and Ansible).
With vendoring, you'll also run into problems in distro packaging ecosystems, in particular, because they patch projects and de-vendor their deps. For example, if Fedora has a dependency that is already packaged separately, they'd mark it as your dependency, remove your vendored code and apply patches so that your code imports that.
What I suggested above is a common and well-understood way of having shims for features in stdlib with backports available.

PR #778, opened 2021-11-22
Declare project URLs in Python package metadata

webknjaz:


setup.cfg
```diff
@@ -2,6 +2,16 @@
...
 url = https://github.com/abhinavsingh/proxy.py
+project_urls =
+  Container Image: DockerHub = https://hub.docker.com/r/abhinavsingh/proxy.py
+  Chat: Gitter = https://gitter.im/proxy.py/community
+  Coverage: Codecov = https://codecov.io/github/abhinavsingh/proxy.py
+  CI: GitHub = https://github.com/abhinavsingh/proxy.py/actions
+  Docs: Changelog = https://proxypy.rtfd.io/en/latest/#changelog
+  Docs: RTD = https://proxypy.rtfd.io/en/latest/
+  Docs: Intro = https://abhinavsingh.com/proxy-py-a-lightweight-single-file-http-proxy-server-in-python/
+  GitHub: issues = https://github.com/abhinavsingh/proxy.py/issues
```
webknjaz, line 13: @abhinavsingh I strongly recommend enabling GitHub Discussions for the repo. This way, you could have a space for googlable StackOverflow-like Q&A and community discussions that would not disrupt the development efforts. And the issues would be kept for tracking bugs and feature requests. I've also seen some projects with routing all the new posts from the users to the Discussions space with the manual creation of the tracked issues. Example: https://github.com/pradyunsg/furo/issues/new/choose.
```suggestion
  GitHub: discussions = https://github.com/abhinavsingh/proxy.py/discussions
  GitHub: issues = https://github.com/abhinavsingh/proxy.py/issues
```

abhinavsingh: Enable it again.  I had it enable, but for some reasons I was finding it hard to track discussions across discussions and issues, so I disabled it.  Lets try again :)

webknjaz: It depends on whether you're watching all of the notifications on the repo or not. Also, it's a place that you don't need to track as actively. It's a place where you'd build a community where other people would help each other.

PR #782, opened 2021-11-23
Publish a GitHub Release after tagging

webknjaz:


.github/workflows/test-library.yml
```diff
@@ -772,4 +772,38 @@ jobs:
...
+        name: python-package-distributions
+        path: dist/
+
+    - name: >-
+        Publish a GitHub Release for
+        ${{ needs.pre-setup.outputs.git-tag }}
+      uses: softprops/action-gh-release@v1
+      with:
+        # body/body_path:  # FIXME: Use once Towncrier is integrated.
+        discussion_category_name: Announcements
```
webknjaz, line 801: @abhinavsingh this is why I asked you to create the discussions category. It'll auto-create announcement topics on release so that there's a centralized place for people to comment on a given version post-release.

webknjaz:
@abhinavsingh this is one of those things that will need to be tested with the alpha release too. It won't run unless a release is triggered and approved.

.github/workflows/test-library.yml
abhinavsingh: Awesome, I haven't checked on the release flows yet . Good to see.  Was thinking of release this week, but might only be able to make one over the weekend.  Will keep ya updated or anyways it will post to announcements :D

webknjaz: @abhinavsingh if you want, we could plan to do this via a video call together. LMK if you'd want to do this and the TZ you're currently in.

PR #814, opened 2021-11-26
update FilterByUpstreamHostPlugin documentation

make-ing:
Update the documentation of the FilterByUpstreamHostPlugin to the new domains which are filtered by default. They are no longer facebook.com but instead google.com


make-ing:
sorry for the stupid PR... i jsut saw that the default branch is develop which actually realy by default filteres the facebook url.

PR #815, opened 2021-11-27
Fix matching requested version with Git tag in GHA

webknjaz:


.github/workflows/test-library.yml
```diff
@@ -270,9 +270,7 @@ jobs:
     - name: >-
         Tag the release in the local Git repo
         as ${{ needs.pre-setup.outputs.git-tag }}
-        for setuptools-scm to se the desired version
```
webknjaz, line 273: There was a typo in "set"

.github/workflows/test-library.yml
```diff
@@ -783,7 +781,10 @@ jobs:
         Tag the release in the local Git repo
         as v${{ needs.pre-setup.outputs.git-tag }}
       run: >-
-        git tag '${{ needs.pre-setup.outputs.git-tag }}'
+        git tag
+        -m '${{ needs.pre-setup.outputs.git-tag }}'
```
webknjaz, line 785: This is updated to produce an annotated tag.

webknjaz:
@abhinavsingh the CI passed so this is ready now.

PR #818, opened 2021-11-28
[GHA] Add container integration test & publish containers to GHCR

abhinavsingh:
TODO:
- Docker step must push all arch images to GHCR
- For dev builds, upon check, push images to Docker Hub (no `latest` tag)
- For stable builds, upon check & approval, push images to Docker Hub (with `latest` tag)
- Check on Docker versioning system for alpha, beta, rc


.github/workflows/test-library.yml
```diff
@@ -680,13 +679,16 @@ jobs:
         docker buildx inspect
         docker buildx ls
     - name: Set PROXYPY_CONTAINER_VERSION
-      run: |
-        echo "PROXYPY_CONTAINER_VERSION=$(echo '${{ needs.pre-setup.outputs.dist-version }}' | tr + .)" > $GITHUB_ENV
+      run: >-
+        echo "PROXYPY_CONTAINER_VERSION=
+          $(echo '${{ needs.pre-setup.outputs.dist-version }}' | tr + .)"
```
webknjaz, line 684: This will probably be problematic because the parser converts the line feeds to spaces in this case. So you'll get `var= <value>` — not sure if that space is safe. I like doing it a bit differently by breaking lines in less essential places:
```suggestion
        echo "PROXYPY_CONTAINER_VERSION=$(
          echo '${{ needs.pre-setup.outputs.dist-version }}' | tr + .
        )"
```

.github/workflows/test-library.yml
```diff
@@ -680,13 +679,16 @@ jobs:
...
+          > $GITHUB_ENV
     - name: Build container
-      run: |
-        make container-buildx \
-          -e PROXYPY_PKG_PATH='dist/${{ needs.pre-setup.outputs.wheel-artifact-name }}' \
-          -e BUILDX_TARGET_PLATFORM='${{ matrix.targetplatform }}' \
+      run: >-
+        make container-buildx
+          -e PROXYPY_PKG_PATH=
+            'dist/${{ needs.pre-setup.outputs.wheel-artifact-name }}'
```
webknjaz, line 690: This concern is the same as above: you'll have the artifact name as a separate argument like `-e var= another_arg` meaning that `$var` will be set to an empty string. But if you break inside of the GHA substitution, it should be fine:
```suggestion
          -e PROXYPY_PKG_PATH='dist/${{
            needs.pre-setup.outputs.wheel-artifact-name
          }}'
```

.github/workflows/test-library.yml
```diff
@@ -195,6 +194,11 @@ jobs:
             && github.event.inputs.release-version
             || steps.scm-version.outputs.dist-version
         }}-py3-none-any.whl')
+    - name: Set PROXYPY_CONTAINER_VERSION env
+      shell: bash
+      run: >-
+        VER=$(echo '${{ needs.pre-setup.outputs.dist-version }}' | tr + .);
```
webknjaz, line 200: This won't work because `needs` refers to the jobs that you've listed in the `needs:` setting of the current job. And you cannot refer to the outputs before they are created. They will be filled out after this job finishes. Those values are taken from the steps so you can refer to the previous steps by their ids using the `steps` context, the same way the mappings to the outputs are set up later.

.github/workflows/test-library.yml
```diff
@@ -195,6 +194,11 @@ jobs:
...
+        echo "PROXYPY_CONTAINER_VERSION=$VER" > $GITHUB_ENV
```
webknjaz, line 201: Since you're already in the "calculation job", use the same pattern and set the outputs. You could copy-paste what's in the artifact-name job:
```suggestion
    - name: Set PROXYPY_CONTAINER_VERSION env
      shell: bash
      run: >-
        VER=$(echo '${{
            steps.request-check.outputs.release-requested == 'true'
            && github.event.inputs.release-version
            || steps.scm-version.outputs.dist-version
        }}' | tr + .);
        echo "::set-output name=container-version::$VER"
```

Then, add a mapping to the `outputs:` section and you'll be able to use it in other jobs via the `needs` context.

.github/workflows/test-library.yml
```diff
@@ -195,6 +195,12 @@ jobs:
             && github.event.inputs.release-version
             || steps.scm-version.outputs.dist-version
         }}-py3-none-any.whl')
+    - name: Set container version
+      id: container-version
+      shell: bash
+      run: >-
+        VER=$(echo '${{ needs.pre-setup.outputs.dist-version }}' | tr + .);
+        print('::set-output name=version::$VER')
```
webknjaz, line 203: This is copy-pasted from the Python block. Convert everything to Python or switch the syntax here to bash:
```suggestion
        echo "::set-output name=version::$VER"
```

.github/workflows/test-library.yml
abhinavsingh: Gotha, makes sense, thank you :)

.github/workflows/test-library.yml
```diff
@@ -679,15 +685,21 @@ jobs:
...
+            '${{ needs.pre-setup.outputs.container-version }}'
+    - name: Run & Test container
       run: |
-        make container-buildx \
-          -e PROXYPY_PKG_PATH='dist/${{ needs.pre-setup.outputs.wheel-artifact-name }}' \
-          -e BUILDX_TARGET_PLATFORM='${{ matrix.targetplatform }}' \
-          -e PROXYPY_CONTAINER_VERSION='${{ env.PROXYPY_CONTAINER_VERSION }}'
+        docker run\
+          -p 8899:8899\
+          abhinavsingh/proxy.py:\
```
webknjaz, line 700: This will inject a space and you'll get `container-name: other-version-arg`

webknjaz (dismissed):

.github/workflows/test-library.yml
```diff
@@ -679,15 +689,25 @@ jobs:
...
-          -e PROXYPY_CONTAINER_VERSION='${{ env.PROXYPY_CONTAINER_VERSION }}'
+    - name: Build container & Push to GHCR
+      run: >-
+        docker images;
+        make container-buildx
+        -e PROXYPY_PKG_PATH='dist/${{
+          needs.pre-setup.outputs.wheel-artifact-name
+        }}'
+        -e BUILDX_TARGET_PLATFORM='${{ matrix.targetplatform }}'
+        -e PROXYPY_CONTAINER_VERSION='${{
```
webknjaz, line 700: In case you want to have a shared image on GHCR, here's an example of having a unique workflow run based tag to be reused across jobs: https://github.com/ansible/creator-ee/blob/8f2ef3e/.github/workflows/release.yml#L114-L118.

abhinavsingh: Thank you, can borrow code example from here.  Here we'll need one image / arch / workflow.  Then push all arch images (with common tag) to registries.

abhinavsingh: Idea is to get this result (similar to `python-alpine`)

Screen Shot 2021-11-29 at 1 37 56 AM

Currently we are here

Screen Shot 2021-11-29 at 1 44 31 AM

abhinavsingh: This PR now pushes to GHCR.  Containers available here https://github.com/abhinavsingh/proxy.py/pkgs/container/proxy.py

But looks like GHCR doesn't support list manifests and we can't get something like DockerHub on GHCR.  Will dig more later as part of a separate PR.  Have to get back eventually to push to DockerHub which I have skipped in this PR.

.github/workflows/test-library.yml
```diff
@@ -709,15 +716,89 @@ jobs:
...
+        }}
+        --hostname 0.0.0.0
+        --enable-web-server
+        --local-executor && ./tests/integration/test_integration.sh 8899
+    - name: Login to GHCR
+      uses: docker/login-action@v1
+      with:
+        registry: ghcr.io
+        username: ${{ github.actor }}
+        password: ${{ secrets.GITHUB_TOKEN }}
```
webknjaz, line 747: This context is only available for the event triggers that are safe and don't have a high risk of running untrusted code contributed by random people via PRs.

.github/workflows/test-library.yml
webknjaz: FWIW I only use GHCR as a staging area before publishing to more "official" registries, I don't think it's meant to be a full replacement of those.

.github/workflows/test-library.yml
abhinavsingh: +1 , also updated repo settings to only allow whitelisted actions.

PR #972, opened 2022-01-12
Bail out early for non-HTTP but HTTP looking protocols

abhinavsingh:
- Add support in `Url` to parse all types of schemes.
- `HttpProtocolHandler` will bail out if parsed scheme from request line is not `http`.  We don't even expect `https`, as for HTTPS request proxies, `CONNECT` is used anyways, and not a `https` URL in request line.

TL;DR -- `HttpProtocolHandler` will only accept web and proxy server requests


abhinavsingh:
@webknjaz Wondering if there is a workaround.  What makes this worse is that `make lib-lint` doesn't fails locally.  But it does on GHA.  Sphinx doc running into warnings while it shouldn't , related Sphinx bug here https://github.com/sphinx-doc/sphinx/issues/4961. Opened since 2018.

I am wondering if we'll have to live with a red CI now :) (in case of a missing workaround).

webknjaz:
I see two problems here:
1) The new (released earlier today or yesterday) version of Sphinx started to emit warnings for hardcoded links that have matching extlink/role equivalents. This seems to not work under macOS according to the bug on the upstream tracker I saw today. Pinning the dependency versions properly should help avoid sudden dependency upgrades with surprises.
2) There are also warnings for ambiguous references, you should fix them by using the proper/full notation.

proxy/http/url.py
```diff
@@ -68,29 +68,41 @@ def from_bytes(cls, raw: bytes) -> 'Url':
         For a HTTPS connect tunnel, url is like ``httpbin.org:443``
         For a HTTP proxy request, url is like ``http://httpbin.org/get``
 
+        proxy.py internally never expects a https scheme in the request line.
+        But `Url` class provides support for parsing any scheme present in the URLs.
```
webknjaz, line 72: This is not inline code but a reference.

proxy/http/url.py
```diff
@@ -68,29 +68,41 @@ def from_bytes(cls, raw: bytes) -> 'Url':
...
+        e.g. ftp, icap etc.
+
+        If a url with no scheme is parsed, e.g. ``//host/abc.js``, then scheme
+        defaults to `http`.
```
webknjaz, line 76: Same here, reference syntax used, but you probably wanted it to be rst.

abhinavsingh:
> 1. Pinning the dependency versions properly should help avoid sudden dependency upgrades with surprises.

Thank you.  Probably pinning is the right way to approach instead of fighting it immediately.  I hope should be resolved in future releases.

Good point about also updating bad references.

webknjaz:
FYI the deps are only unpinned for the spelling env and pinned for others. This is why only one invocation is failing.

PR #1180, opened 2022-06-05
Add support for dynamic reverse proxy routing

### LINKED ISSUE #1132: Make it possible to have a dynamic reverse proxy

abhinavsingh:
Plugins must implement the `def handle_route(self, request: HttpParser, pattern: re.Pattern[Any]) -> Url`
method from the base class to provide a choice of upstream url to serve.

This work is backward compatible a.k.a. Old style static route patterns
with hardcoded upstream urls will continue to work as-is.

See `ReverseProxyBasePlugin` documentation for more details.

Fixes #1132
- #1132

### DIFF
--- a/.github/FUNDING.yml
+++ b/.github/FUNDING.yml
@@ -1,3 +1,4 @@
+---
 # These are supported funding model platforms
 
 github: abhinavsingh
--- a/.github/FUNDING.yml
+++ b/.github/FUNDING.yml
@@ -1,0 +1,3 @@
+# These are supported funding model platforms
+
+github: abhinavsingh
--- a/.github/workflows/test-library.yml
+++ b/.github/workflows/test-library.yml
@@ -72,7 +72,7 @@ jobs:
       container-platforms: ${{ steps.container.outputs.platforms }}
     steps:
     - name: Switch to using Python 3.9 by default
-      uses: actions/setup-python@v3
+      uses: actions/setup-python@v4
       with:
         python-version: 3.9
     - name: >-
@@ -135,7 +135,7 @@ jobs:
     - name: Set up pip cache
       if: >-
         steps.request-check.outputs.release-requested != 'true'
-      uses: actions/cache@v3.0.2
+      uses: actions/cache@v3.0.4
       with:
         path: ${{ steps.pip-cache-dir.outputs.dir }}
         key: >-
@@ -225,7 +225,7 @@ jobs:
 
     steps:
     - name: Switch to using Python v3.10
-      uses: actions/setup-python@v3
+      uses: actions/setup-python@v4
       with:
         python-version: '3.10'
     - name: >-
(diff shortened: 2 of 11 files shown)


abhinavsingh:
@piraz @rpgmaster280 Please take a look at this PR.  I have updated the example plugin with a dynamic route implementation.  You'll need to write a plugin to handle your own custom dynamic routes.  Best

PR #1329, opened 2023-04-25
Added authorization via LDAP

svesh:


requirements-release.txt
```diff
@@ -1,2 +1,3 @@
 setuptools-scm == 6.3.2
 twine==3.8.0
+python-ldap==3.4.3
```
abhinavsingh, line 3: 1) We should not add this as a release requirement
2) Since `auth_ldap` is just an example plugin, it's upto user to install the dependency if they choose to use this plugin
3) We kind of do similar thing for other plugins where a dependency might be required.

Otherwise, now to install `proxy.py`, users will also need to install `python-ldap` even when not using the plugin.

abhinavsingh (commented):
Thank you for the PR.  Good to merge once we can remove `python-ldap` as a mandatory dependency.

PR #1345, opened 2023-08-24
Migrate away from setuptools_scm_git_archive

tjni:
I'm packaging this in [nixpkgs](https://github.com/NixOS/nixpkgs) and would like to avoid a dependency on setuptools_scm_git_archive. I followed the [migration instructions](https://pypi.org/project/setuptools-scm-git-archive/) and picked >= 7.0.5 because I saw that there was a bug reported in https://github.com/pypa/setuptools_scm/issues/745 that was fixed in that version.

I do not know how to test this so help here is appreciated.


abhinavsingh:
@tjni We may need to look a little deeper to remove this dependency.  At least, as of now, our code tries to import `proxy.common._scm_version` and absence of this file will lead to errors.   In fact, this integration was a contribution by @webknjaz , adding him for more insights here.

webknjaz:
The change is correct. The archival file should provide more metadata that newer versions of setuptools-scm use. As long as you don't need to support ancient Python versions that setuptools-scm no longer supports, this is good to go.

abhinavsingh:
@webknjaz Thank you for your advice on this :). @tjni Back then workflows were broken and I had little bandwidth to invest into it.  Today I have updated the `develop` branch so that workflows are stable again.  Updated your branch, will try and merge once after workflow passes.  Thank you folks.

abhinavsingh:
@tjni jFYI, I had to add back `"setuptools-scm[toml]>=6,!=7.0.0,!=7.0.1,!=7.0.2"` today because `"setuptools-scm[toml]>=7.0.5"` doesn't seem to support `Python 3.6` https://pypi.org/project/setuptools-scm/7.0.5/

I am not sure why this never triggered an error before till now.   But, workflow failures started happening in this PR
- https://github.com/abhinavsingh/proxy.py/pull/1452

@webknjaz Wondering if you observe any typo/glaring issue in the PR.  It simply migrates away from `requirements*` files to setup.cfg `extra_require` configuration.

PR #1347, opened 2023-09-20
Remove use of ssl.wrap_socket

s-t-e-v-e-n-k:
ssl.wrap_socket() has been deprecated since Python 3.7, and isn't recommended for use, and further, has been removed in Python 3.12. ssl.SSLContext().wrap_socket() is the new path forward, so switch the one callsite and the two test cases to use it instead.


abhinavsingh:
@s-t-e-v-e-n-k Thanks for the patch, apologies for no show on GitHub for months.  Will include this in the new release this month.

abhinavsingh:
@s-t-e-v-e-n-k Thank you for the PR, and apologies for delaying the merge, was away from OSS for long.  I have restarted the workflows and shall merge once after they passed.  Best

abhinavsingh:
@s-t-e-v-e-n-k https://results.pre-commit.ci/run/github/12228178/1712825888.cSxDle-WSWiDTFOmoJ9xpw

```
proxy/core/connection/client.py: note: In member "wrap" of class "TcpClientConnection":
proxy/core/connection/client.py:46:22: error: Unexpected keyword argument
"certfile" for "wrap_socket" of "SSLContext"  [call-arg]
            self._conn = ssl_context.wrap_socket(
                         ^
/pc/clone/k6zKvJ0aTyeNk_FxLD3xpg/py_env-python3/lib/python3.11/site-packages/mypy/typeshed/stdlib/ssl.pyi:362:9: note: "wrap_socket" of "SSLContext" defined here
proxy/core/connection/client.py:46:22: error: Unexpected keyword argument
"keyfile" for "wrap_socket" of "SSLContext"  [call-arg]
            self._conn = ssl_context.wrap_socket(
                         ^
/pc/clone/k6zKvJ0aTyeNk_FxLD3xpg/py_env-python3/lib/python3.11/site-packages/mypy/typeshed/stdlib/ssl.pyi:362:9: note: "wrap_socket" of "SSLContext" defined here
proxy/core/connection/client.py:46:22: error: Unexpected keyword argument
"ssl_version" for "wrap_socket" of "SSLContext"  [call-arg]
            self._conn = ssl_context.wrap_socket(
                         ^
/pc/clone/k6zKvJ0aTyeNk_FxLD3xpg/py_env-python3/lib/python3.11/site-packages/mypy/typeshed/stdlib/ssl.pyi:362:9: note: "wrap_socket" of "SSLContext" defined here
Found 3 errors in 1 file (checked 218 source files)
```

abhinavsingh:
@s-t-e-v-e-n-k 

>> ssl.wrap_socket() has been deprecated since Python 3.7

- We still support Python 3.6 which is important for some of the repositories using `proxy.py`
- Looks like these changes breaks thing for 3.6

s-t-e-v-e-n-k:
> * We still support Python 3.6 which is important for some of the repositories using `proxy.py`
> * Looks like these changes breaks thing for 3.6

I *think* 3.6 should be fine for these changes, based on my reading of https://docs.python.org/3/library/ssl.html -- except I was assuming `SSLContext.wrap_socket` took the same arguments as `ssl.wrap_socket`, which it does not. Bad me for not reading the docs closely enough! I'm rebasing my branch as well as correcting the above. I'll test on a Python 3.6 system once I get my hands on one.

abhinavsingh:
> > * We still support Python 3.6 which is important for some of the repositories using `proxy.py`
> > * Looks like these changes breaks thing for 3.6
> 
> I _think_ 3.6 should be fine for these changes, based on my reading of https://docs.python.org/3/library/ssl.html -- except I was assuming `SSLContext.wrap_socket` took the same arguments as `ssl.wrap_socket`, which it does not. Bad me for not reading the docs closely enough! I'm rebasing my branch as well as correcting the above. I'll test on a Python 3.6 system once I get my hands on one.

I think a straight forward way would be to:
1) Keep using old and in-secure wrap_socket
2) For other versions we can adopt the new paradigm

I was just looking to support Python 3.12, but looks like `wrap_socket` is now removed in 3.12 :D

abhinavsingh:
Superseded by https://github.com/abhinavsingh/proxy.py/pull/1443

PR #1350, opened 2023-09-27
Update connection.py - Remove unused log line

wangzhe3224:


abhinavsingh:
@wangzhe3224 Thank you, but this is just there for development purposes when dev wants to spit out raw packet for inspection during development.

Issue #1361, opened 2024-02-28
Proxy to download http file slow


YK-Samgo:
**Describe the bug**
Proxy to download http file slow

**To Reproduce**
1. Run `proxy`
2. Download one big file on local server with and without proxy. The speed diffs a lot.
3. I can see the `proxy` process used up one whole core in `top`

Download without proxy. Everything finished in one second on my server.

```shell
sudo curl -v http://127.0.0.1/api/bigfile -o /dev/null
*   Trying 127.0.0.1:80...
  % Total    % Received % Xferd  Average Speed   Time    Time     Time  Current
                                 Dload  Upload   Total   Spent    Left  Speed
  0     0    0     0    0     0      0      0 --:--:-- --:--:-- --:--:--     0* Connected to 127.0.0.1 (127.0.0.1) port 80 (£0)
> GET /api/bigfile HTTP/1.1
> Host: 127.0.0.1
> User-Agent: curl/7.74.0
> Accept: */*
>
* Mark bundle as not supporting multiuse
< HTTP/1.1 200 OK
< Date: Wed, 28 Feb 2024 09:25:32 GMT
< Server: Apache/2.4.56 (Debian)
< Upgrade: h2,h2c
< Connection: Upgrade
< Last-Modified: Wed, 28 Feb 2024 09:25:24 GMT
< ETag: "100000000-6126db8fc99f6"
< Accept-Ranges: bytes
< Content-Length: 4294967296
<
{ [32513 bytes data]
100 4096M  100 4096M    0     0  3444M      0  0:00:01  0:00:01 --:--:-- 3444M
```

Download with proxy. I can't wait it to complete.

```shell
sudo curl -v http://127.0.0.1/api/bigfile -o /dev/null -x 'http://127.0.0.1:8899'
*   Trying 127.0.0.1:8899...
  % Total    % Received % Xferd  Average Speed   Time    Time     Time  Current
                                 Dload  Upload   Total   Spent    Left  Speed
  0     0    0     0    0     0      0      0 --:--:-- --:--:-- --:--:--     0* Connected to 127.0.0.1 (127.0.0.1) port 8899 (£0)
> GET http://127.0.0.1/api/bigfile HTTP/1.1
> Host: 127.0.0.1
> User-Agent: curl/7.74.0
> Accept: */*
> Proxy-Connection: Keep-Alive
>
* Mark bundle as not supporting multiuse
< HTTP/1.1 200 OK
[...]

Issue #498, opened 2021-02-03
[ModifyChunkResponsePlugin] fails with "OSError when flushing buffer to client" with HTTPS PROXY


ar4development:
1. Run the proxy embedded:

```
import proxy
from proxy.plugin.modify_chunk_response import ModifyChunkResponsePlugin

if __name__ == '__main__':
    with proxy.start(
            ['--host', '127.0.0.1',
             '--port', '12233',
             '--log-level', 'DEBUG',
             '--ca-cert-file', '/YOUR_PATH_TO_CERT_FILES/myCA.pem',
             '--ca-key-file', '/YOUR_PATH_TO_CERT_FILES/myCA.key',
             '--ca-signing-key-file', '/YOUR_PATH_TO_CERT_FILES/myCA-signing.key'],
            plugins=
            [b'proxy.plugin.modify_chunk_response.ModifyChunkResponsePlugin',
             proxy.plugin.modify_chunk_response.ModifyChunkResponsePlugin]):
        import time
        time.sleep(10000)
```
2. When the main thread sleeps call the following `curl` command: `curl -v --proxy http://localhost:12233 --cacert /YOUR_PATH_TO_CERT_FILES/myCA.pem https://httpbin.org`

**Expected behavior**
Fake chunks from plugin are received by curl

**Observed behavior**

In curl:

```
*   Trying 127.0.0.1:12233...
* TCP_NODELAY set
* Connected to localhost (127.0.0.1) port 12233 (#0)
* allocate connect buffer!
* Establish HTTP proxy tunnel to httpbin.org:443
> CONNECT httpbin.org:443 HTTP/1.1
> Host: httpbin.org:443
> User-Agent: curl/7.68.0
> Proxy-Connection: Keep-Alive
> 
< HTTP/1.1 200 Connection established
< 
* Proxy replied 200 to CONNECT request
* CONNECT phase completed!
* ALPN, offering h2
* ALPN, offering http/1.1
* successfully set certificate verify locations:
*   CAfile: /home/alexey/Desktop/Dev/testing/CA/myCA.pem
  CApath: /etc/ssl/certs
* TLSv1.3 (OUT), TLS handshake, Client hello (1):
* CONNECT phase completed!
* CONNECT phase completed!
* TLSv1.3 (IN), TLS handshake, Server hello (2):
* TLSv1.3 (IN), TLS handshake, Encrypted Extensions (8):
* TLSv1.3 (IN), TLS handshake, Certificate (11):
[...]

PR #1362, opened 2024-02-28
Wait until all data in buffer is flushed to client

YK-Samgo:
Wait until all data in buffer is flushed to client when upstream server finishes.

This bug is met after #1361

With high performance upstream and large memory, `proxy` will keep a lot of data in buffer, and the upstream link may be closed by upstream server. Then `proxy` will mistake this shutdown as `server tear down` and kill the whole work even though there is still data in buffer.

I think it's more graceful to wait until all data flushed to client in this case.

This is a performance-related and random issue. I think it's also the reason of #498

 In my server:

```
          1000M              10000M+
client ==== proxy.py ==== upstream server
```

May be there is a similar issue when upload large file, but I didn't meet. Or it won't be a problem since client will wait until response.

The bug #1361 will slow down the upstream so it will never be faster than client, which avoid the issue.


abhinavsingh:
@YK-Samgo Thank you for the PR, and apologies for delaying the merge, was away from OSS for long.  I have restarted the workflows and shall merge once after they passed.  Best

abhinavsingh:
@YK-Samgo Thanks a ton for pointing out this issue.  I have cherry-picked your commit into this PR and made further enhancements.  Would appreciate if you can review https://github.com/abhinavsingh/proxy.py/pull/1385

PR #1389, opened 2024-04-18
Ability to override `--data-dir` for scenarios when `proxy.py` is running as a user with no home directory

### LINKED ISSUE #1316: `DEFAULT_DATA_DIRECTORY_PATH` Should be configurable
`DEFAULT_DATA_DIRECTORY_PATH` is hardcoded to `~/.proxy`.

`DEFAULT_DATA_DIRECTORY_PATH` is configurable via an option, for example `--data-directory=/opt/proxy.py/data`

abhinavsingh:
- Fixes https://github.com/abhinavsingh/proxy.py/issues/1316

### DIFF
--- a/.github/workflows/test-library.yml
+++ b/.github/workflows/test-library.yml
@@ -188,7 +188,7 @@ jobs:
     - name: Set the expected dist artifact names
       id: artifact-name
       run: |
-        print('::set-output name=sdist::proxy.py-${{
+        print('::set-output name=sdist::proxy_py-${{
             steps.request-check.outputs.release-requested == 'true'
             && github.event.inputs.release-version
             || steps.scm-version.outputs.dist-version
--- a/.gitignore
+++ b/.gitignore
@@ -36,3 +36,6 @@ build
 
 pyreverse.png
 profile.svg
+
+*-pre-push
+jaxl-api-credentials*.json
--- a/.vscode/settings.json
+++ b/.vscode/settings.json
@@ -19,7 +19,7 @@
   "typescript.preferences.quoteStyle": "single",
   "[python]": {
     "editor.wordBasedSuggestions": "matchingDocuments",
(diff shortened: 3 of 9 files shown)


.github/workflows/test-library.yml
```diff
@@ -188,7 +188,7 @@ jobs:
     - name: Set the expected dist artifact names
       id: artifact-name
       run: |
-        print('::set-output name=sdist::proxy.py-${{
+        print('::set-output name=sdist::proxy_py-${{
```
abhinavsingh, line 191: @webknjaz Apologies for tagging you again, but wanted to highlight this change as I am unsure what upstream changes might have caused this.  Somehow generated `.tar.gz` name has suddenly changed

webknjaz: @abhinavsingh no need to apologize :) This update has been widely noticed across many projects lately. `setuptools` made updates to the sdist base name normalization code in one of the recent releases which resulted in underscores appearing in place of `.` or `-`. This would've been more stable if `setuptools` was pinned for the build env (I don't think I implemented this in my earlier PRs but I tend to do this in my other projects).

Here's the corresponding change log entry upstream: https://setuptools.pypa.io/en/latest/history.html#v69-3-0.

webknjaz: @abhinavsingh while I'm here, I also noticed the use of `::set-output` GHA workflow command syntax which has been deprecated for a while. The new way is appending to a file path provided to by the `${GITHUB_OUTPUT}` environment variable. For example: https://github.com/cherrypy/cheroot/blob/3591a1c/.github/workflows/ci-cd.yml#L309-L327.

abhinavsingh: @webknjaz You are a charm.  Thank you for providing links and insights, I'll dig into them 🙏

PR #1400, opened 2024-04-26
Fix ssh tunnel use case

abhinavsingh:


abhinavsingh:
Just discovered https://github.com/actions/setup-python/issues/850#issuecomment-2072657610 :). Will fix later, merging this PR due to known upstream reasons for certain failures.

TL;DR -- GHA macOS-latest now defaults to Apple Silicon on which certain python and node versions are not available.

PR #1401, opened 2024-04-29
Disable mandatory TLS verification with --insecure

d4xfe:
`getpeercert()` only returns the certificate with enabled certificate validation. Because of this, I updated it to `getpeercert(True)`, which will return a binary blob of the servers certificate regardless of the verification mode.

I added the util method `cert_der_to_dict` using `ssl` internals to convert the blob to a python dict like `getpeercert()` would.
The flag `--insecure` was introduced to disable certificate validation of ssl sockets.
This allows TLS interception when the server is using a self-signed certificate.

This is my first pull request on GitHub and I'm still a bit confused. So if I should change anything let me know.


abhinavsingh:
@d4xfe Thank you, looks excellent considering this is your first PR.  Congratulations.

Can you please:
1) Resolve conflicts
2) Sync your branch with `develop`.   For some reasons, I see diffs in your PR which are not part of your changes

d4xfe:
@abhinavsingh Thanks. I've synced the branches and I don't conflicts anymore.

abhinavsingh:
@d4xfe Thank you for this work and the PR.  There are some workflow issues.  I am going create a new PR based on top of your contribution and try to this ship.

PR #1414, opened 2024-06-01
[WIP] Starting with metrics and prometheus 

### LINKED ISSUE #1320: Support Prometheus via plugin
Sending stats and access logs to https://prometheus.io/ would be a great addition. For inspiration, there's https://github.com/gmm42/tinyproxy_exporter/blob/master/tinyproxy_exporter

magmax:
Fixes #1320

**WORK IN PROGRESS**

It is not fully functional right now. First of all, it seems that my local environment doesn't accomplish all the requirements to run the pre-commit and pre-push hooks, but I wanted to share my current code with you.

The idea is:
1. Plugins have the ability to send metrics, but they do not know anything about prometheus.
2. Prometheus plugin receives the message and process it.

Problems:
- prometheus_client has a singleton `REGISTRY`.
- it is not possible to register two metrics with same name in the `REGISTRY`.
- but I didn't want to use `prometheus_client.REGISTRY` on core.


abhinavsingh:
@magmax Thanks for the initial PR.  I took a deep look and it required a different strategy to work w/ `proxy.py`.  Please check PR https://github.com/abhinavsingh/proxy.py/pull/1447 which currently is in working state.

PR #1420, opened 2024-06-09
Add `ModifyRequestHeaderPlugin`

abhinavsingh:


proxy/plugin/modify_request_header.py
```diff
@@ -0,0 +1,31 @@
...
+
+
+class ModifyRequestHeaderPlugin(HttpProxyBasePlugin):
+    """Modify request header before sending to upstream server."""
+
+    def handle_client_request(self, request: HttpParser) -> Optional[HttpParser]:
+        """NOTE: Modification of client request under TLS interception WILL NOT WORK
+        through before_upstream_connection hook override.  This example plugin is demonstrate
+        how to modify headers for all scenarios (with or without interception).
+        """
```
abhinavsingh: @sandrich Please try this, wrote a little note why `before_upstream_connection` won't work for request modification when under TLS interception.

`before_upstream_connection` is mostly for overriding/customising upstreams.   They also work as request modifiers when not under TLS interception.   But when interception is ON, `handle_client_request` is the correct way for modifying requests.

PR #1452, opened 2024-08-12
`setup.cfg/extras_require` configuration

abhinavsingh:


webknjaz:
@abhinavsingh that's probably because you dropped the lockfiles, including the one that pinned the build env, providing reproducible builds. Those were supposed to be updated periodically so that you'd see the breakage in corresponding bump PRs.

Abusing extras as environment definitions is discouraged as it's misuse of their actual purpose. Therefore I'm waiting for the dependency groups and lockfiles PEPs that would standardize the corresponding mechanisms correctly. I see you dropped reproducibility in favor of exposing `pip install proxy.py[testing]` to the end-users through packaging metadata.

PR #1535, opened 2025-11-25
Feature: 'Forward All' flag for ProxyPool plugin.

Adnan-Odimah:
Added a flag '--forward-all' that works alongside the ProxyPool plugin to forward all requests to the remote proxy including private requests.


Adnan-Odimah:
@abhinavsingh, can't seem to get the precommit tests to pass. Keep getting:
```
proxy/__main__.py

  3:1      RST399 Missing matching underlin
```
Could I get some help with this?

